PR #5, opened 2022-12-20
fix(naming): Rename variant to implementation

### LINKED ISSUE #3: rename variant to implementation
The current code refers to "variants" in a number of places, typically when talking about multiple implementations of the same function (e.g. add<int32,int32> and add<int64,int64>).  Since the spec calls these "implementations" and the spec also has a different concept that it calls "variant" I think it would be good to update the repo changing all instances of "variant" to "implementation".

chaojun-zhang:
Closes #3

### DIFF
--- a/include/substrait/function/Extension.h
+++ b/include/substrait/function/Extension.h
@@ -43,7 +43,7 @@ class Extension {
   /// Add a scalar function implementation.
   void addScalarFunctionImpl(const FunctionImplementationPtr& functionImpl);
 
-  /// Add a aggregate function implementation.
+  /// Add an aggregate function implementation.
   void addAggregateFunctionImpl(const FunctionImplementationPtr& functionImpl);
 
   /// Add a window function implementation.
--- a/include/substrait/function/Extension.h
+++ b/include/substrait/function/Extension.h
@@ -20,7 +20,7 @@ struct TypeVariant {
 
 using TypeVariantPtr = std::shared_ptr<TypeVariant>;
 
-using FunctionVariantMap =
+using FunctionImplMap =
     std::unordered_map<std::string, std::vector<FunctionImplementationPtr>>;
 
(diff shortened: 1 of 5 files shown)


include/substrait/function/Function.h
```diff
@@ -95,7 +95,7 @@ struct FunctionImplementation {
   ParameterizedTypePtr returnType;
   std::optional<FunctionVariadic> variadic;
 
-  /// Test if the actual types matched with this function variant.
+  /// Test if the actual types matched with this function implement.
```
EpsilonPrime, line 98: It's longer but would this read better if it said "this function's implementation"?

EpsilonPrime (approved):

include/substrait/function/Function.h
chaojun-zhang: Rebased and updated

include/substrait/function/Extension.h
```diff
@@ -40,40 +40,40 @@ class Extension {
...
       const std::vector<std::string>& extensionFiles);
 
-  /// Add a scalar function variant.
-  void addScalarFunctionVariant(const FunctionImplementationPtr& functionVariant);
+  /// Add a scalar function implementation.
+  void addScalarFunctionImpl(const FunctionImplementationPtr& functionImpl);
 
-  /// Add a aggregate function variant.
-  void addAggregateFunctionVariant(const FunctionImplementationPtr& functionVariant);
+  /// Add a aggregate function implementation.
```
westonpace, line 46:
```suggestion
  /// Add an aggregate function implementation.
```

include/substrait/function/Function.h
westonpace, line 98:
```suggestion
  /// Test if the actual types matched with this function's implementation.
```

westonpace (approved):
This is great, thanks.  Two minor wording nits (I can apply them myself and will merge).

Issue #10, opened 2023-01-31
Add library for converting binary and text plans


EpsilonPrime:
As part of the proposal to add a new text-based language for representing plans add a library that can be used to convert from one to another.  Ultimately this new text-based language will replace the JSON-encoded protobuf used today.

EpsilonPrime:
take

EpsilonPrime:
Done!

PR #11, opened 2023-01-31
GH-10: [C++] Add C++ protobuf library for substrait plans

EpsilonPrime:
Adds a C++ protobuf library over third_party/substrait/proto for use in manipulating plans.


chaojun-zhang (approved):
LGTM

westonpace:
How much do we want to build 3rd party dependencies from source vs. expect developers to obtain a copy on their own.  For example, I updated the CI to get protobuf from apt.  For CI testing it is very valuable to be able to rely on apt for packages because GHA maintains an apt mirror and can download and install the package very quickly.

This PR appears to be adding the ability to build protobuf from source.  Although, it seems to be doing it in a way that requires we build protobuf from source (I think our current fmt, etc. do this as well).

Could we perhaps setup the cmake so that it only builds the library from source if find_package fails?  This is roughly how Arrow does things.

EpsilonPrime:
> This PR appears to be adding the ability to build protobuf from source. Although, it seems to be doing it in a way that requires we build protobuf from source (I think our current fmt, etc. do this as well).

This PR adds the functionality to convert the substrait protos into C++ and compile them (providing a library so those symbols are available to C++).  The substrait repo is included as a submodule already in substrait-cpp and is presumably that way so we can ensure that the versions are matched.  The protobuf compiler package is only referenced with a find_package rule which is consistent with your CI effort.

westonpace:
> This PR adds the functionality to convert the substrait protos into C++ and compile them (providing a library so those symbols are available to C++). The substrait repo is included as a submodule already in substrait-cpp and is presumably that way so we can ensure that the versions are matched. The protobuf compiler package is only referenced with a find_package rule which is consistent with your CI effort.

Ah, sorry, I definitely scanned this poorly.  Looks like CI is failing then because it's unable to find protoc.  Can you add protoc to the setup-ubuntu script?

This makes sense now and I agree we want to generate these files.

EpsilonPrime:
> Ah, sorry, I definitely scanned this poorly. Looks like CI is failing then because it's unable to find protoc. Can you add protoc to the setup-ubuntu script?

Added as separate PR #20 as I believe the CI change has to be already in.

westonpace:
> Added as separate PR https://github.com/substrait-io/substrait-cpp/pull/20 as I believe the CI change has to be already in.

I don't think it does.  I believe the CI pulls the workflow yml file from `main` but the checkout (and thus the `setup.sh` script) will be from the PR.

Either way, I've merged #20.  Do you want to push an empty commit to trigger CI again?

westonpace:
Nvm, I can rerun it

westonpace:
Still didn't install it.  Perhaps a rebase is needed?

EpsilonPrime:
The rebase did the trick.  Thanks!

substrait/proto/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
+# SPDX-License-Identifier: Apache-2.0
+
+find_package(Protobuf REQUIRED)
+include_directories(${Protobuf_INCLUDE_DIRS})
+include_directories(${CMAKE_CURRENT_BINARY_DIR})
+
+
+SET(PROTO_INCLUDE_DIR "${CMAKE_SOURCE_DIR}/third_party/substrait/proto/substrait")
+cmake_path(GET PROTO_INCLUDE_DIR PARENT_PATH PROTO_INCLUDE_TOPLEVEL_DIR)
+SET(PROTOBUF_IMPORT_DIRS ${PROTO_INCLUDE_DIR}/extensions)
```
westonpace, line 10: Do you have any extensions?

substrait/proto/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+
+    cmake_path(REMOVE_EXTENSION RELATIVE_PROTO_PATH)
+
+    set(PROTO_HDR ${PROTO_OUTPUT_PARENT_DIR}/${RELATIVE_PROTO_PATH}.pb.h)
+    set(PROTO_SRC ${PROTO_OUTPUT_PARENT_DIR}/${RELATIVE_PROTO_PATH}.pb.cc)
+    add_custom_command(
+            OUTPUT ${PROTO_SRC} ${PROTO_HDR}
+            COMMAND protobuf::protoc
+            "--proto_path=${PROTO_INCLUDE_TOPLEVEL_DIR}"
+            "--cpp_out=${PROTO_OUTPUT_TOPLEVEL_DIR}"
```
westonpace, line 36: This seems to output to `substrait/plan.pb.h` so you would need `#include "substrait/plan.pb.h"`.  Can we make this output instead to `substrait/proto/plan.pb.h`?

substrait/proto/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+            ${PROTO_FILE}
+            DEPENDS ${PROTO_FILE} protobuf::protoc
+            COMMENT "Generate C++ protocol buffer for ${PROTO_FILE}"
+            VERBATIM)
+    list(APPEND PROTO_HDRS ${PROTO_HDR})
+    list(APPEND PROTO_SRCS ${PROTO_SRC})
+endforeach()
+
+# Add the generated protobuf C++ files to our exported library.
+add_library(proto ${PROTO_SRCS})
```
westonpace, line 46: This will create `libproto.a` which I feel might be a little ambiguous.  Can we name it `substrait_proto`?

westonpace (changes requested):
A few questions

substrait/proto/CMakeLists.txt
EpsilonPrime: I do not.   One of the substrait proto definition files in an extensions subdirectory:

https://github.com/substrait-io/substrait/blob/main/proto/substrait/extensions/extensions.proto

substrait/proto/CMakeLists.txt
EpsilonPrime: Done.

substrait/proto/CMakeLists.txt
EpsilonPrime: I had to modify the generated C++ header and source files for this to work.  The correct way to do this would have been to change the proto's package but that would affect all of the other consumers which we have previously stated we didn't want to do.

westonpace:
CC @chaojun-zhang for a second opinion on manipulating the protobuf headers after generation but I think it is fine.  I wonder if we can use this trick to modify the namespace (to `substrait::proto`) while we are at it?

substrait/proto/update_proto_includes.pl
```diff
@@ -0,0 +1,6 @@
+#!/bin/perl -w
+
```
jbalint, line 2: can we please add a comment about what this does?

EpsilonPrime: Done.

EpsilonPrime:
I've updated the rewrite rules to rewrite the proto definition files to have the alternative packages.  It's a bit more invasive as I needed to also update references to the new package.

chaojun-zhang:
> CC @chaojun-zhang for a second opinion on manipulating the protobuf headers after generation but I think it is fine. I wonder if we can use this trick to modify the namespace (to `substrait::proto`) while we are at it?

Yes. currently we  don't have any substrait protobuf headers included.

westonpace (approved):
Clever fixes.  I'm happy we were able to overcome this location problem.

PR #18, opened 2023-02-08
support decode raw string type to instance of io::substrait::Type

chaojun-zhang:
- support  decode rawType to io::substrait::Type

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,7 @@ find_package(Protobuf REQUIRED)
 include_directories(${PROTOBUF_INCLUDE_DIRS})
 
 add_subdirectory(third_party)
+include_directories(.)
 include_directories(include)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake_modules")
--- a/include/substrait/type/Type.h
+++ b/include/substrait/type/Type.h
@@ -189,7 +189,12 @@ class ParameterizedType {
   /// Deserialize substrait raw type string into Substrait extension  type.
   /// @param rawType - substrait extension raw string type
   static std::shared_ptr<const ParameterizedType> decode(
-      const std::string& rawType);
+      const std::string& rawType){
+    return decode(rawType, true);
+  }
+
+  static std::shared_ptr<const ParameterizedType> decode(
+      const std::string& rawType,bool isParameterized);
 
   [[nodiscard]] const bool& nullable() const {
     return nullable_;
--- a/include/substrait/type/Type.h
+++ b/include/substrait/type/Type.h
@@ -189,12 +189,13 @@ class ParameterizedType {
   /// Deserialize substrait raw type string into Substrait extension  type.
   /// @param rawType - substrait extension raw string type
   static std::shared_ptr<const ParameterizedType> decode(
-      const std::string& rawType){
+      const std::string& rawType) {
     return decode(rawType, true);
   }
 
   static std::shared_ptr<const ParameterizedType> decode(
-      const std::string& rawType,bool isParameterized);
+      const std::string& rawType,
+      bool isParameterized);
 
   [[nodiscard]] const bool& nullable() const {
     return nullable_;
@@ -204,8 +205,15 @@ class ParameterizedType {
       const std::shared_ptr<const ParameterizedType>& type) const {
     return nullable() || nullable() == type->nullable();
   }
-  /// Test type is a Wildcard type or not.
+  /// Test whether a type is a Wildcard type or not, always false for non
+  /// StringLiteral type.
   [[nodiscard]] virtual bool isWildcard() const {
+    return false;
+  }
+
+  /// Test whether a Type is a placeholder type or not, always false for non
+  /// StringLiteral type.
+  [[nodiscard]] virtual bool isPlaceholder() const {
     return false;
   }
 
@@ -398,11 +406,14 @@ class ParameterizedTypeBase : public ParameterizedType {
       : ParameterizedType(nullable) {}
 };
 
-/// A string literal type can present the 'any1'.
+/// A string literal type can present the 'any1' or 'T','P1'.
 class StringLiteral : public ParameterizedTypeBase {
  public:
-  explicit StringLiteral(std::string value)
-      : ParameterizedTypeBase(false), value_(std::move(value)) {}
+  StringLiteral(const std::string& value, bool wildcard, bool placeholder)
+      : ParameterizedTypeBase(false),
+        value_(std::move(value)),
+        wildcard_(wildcard),
+        placeholder_(placeholder) {}
 
   [[nodiscard]] std::string signature() const override {
     return value_;
@@ -417,14 +428,23 @@ class StringLiteral : public ParameterizedTypeBase {
   }
 
   [[nodiscard]] bool isWildcard() const override {
-    return value_.find("any") == 0 || value_ == "T";
+    return wildcard_;
   }
+
+  [[nodiscard]] bool isPlaceholder() const override {
+    return placeholder_;
+  }
+
+  /// Return true if value is a integer, false otherwise.
+  [[nodiscard]] bool isInteger() const;
 
   [[nodiscard]] bool isMatch(
       const std::shared_ptr<const ParameterizedType>& type) const override;
 
  private:
   const std::string value_;
+  const bool wildcard_;
+  const bool placeholder_;
 };
 
 using StringLiteralPtr = std::shared_ptr<const StringLiteral>;
--- a/substrait/common/NumberUtils.h
+++ b/substrait/common/NumberUtils.h
@@ -2,6 +2,7 @@
 
 #pragma once
 
+#include <algorithm>
 #include <string>
 
 namespace io::substrait::common {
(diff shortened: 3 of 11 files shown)


westonpace:
Can you explain a bit what the difference is between `ParamterizedXyz` and `Xyz` (for example, what is the difference between `ParameterizedDecimal` and `Decimal`?

chaojun-zhang:
> Can you explain a bit what the difference is between `ParamterizedXyz` and `Xyz` (for example, what is the difference between `ParameterizedDecimal` and `Decimal`?

Type used inside extension YAML file is `ParameterizedType`while  type used inside a substrait plan is called 'Type'.

The difference is  `ParameterizedDecimal`  in a extension YAML file which can accept either a template parameter  or a primitive int as parameters,  while a type `Decimal` reference in a substrait plan can only accept primitive int type as parameter.

westonpace:
Got it.  We use `Decimal` when reading from Substrait because, at that point, the types have been realized and we should know the exact values of precision and scale.

We use `ParameterizedDecimal` when reading from YAML because, it's possible the function is generic, and it uses template arguments like `P` and `S`.

substrait/type/tests/TypeTest.cpp
```diff
@@ -87,34 +87,64 @@ TEST_F(TypeTest, decodeTest) {
...
+      });
+
   testDecode<ParameterizedDecimal>(
       "decimal<P,S>",
       [](const std::shared_ptr<const ParameterizedDecimal>& typePtr) {
         ASSERT_EQ(typePtr->signature(), "dec<P,S>");
         ASSERT_EQ(typePtr->precision()->value(), "P");
         ASSERT_EQ(typePtr->scale()->value(), "S");
       });
+  testDecode<Decimal,false>(
```
westonpace, line 129: I assume something like:

```
const auto& type = ParameterizedType::decode("dec<P, S>", false)
```

should throw an exception (I think the stoi will throw).  Do we want to test that case?

westonpace (approved):
This looks good.  I wonder if we want to test one or two failure cases?  Let me know what you'd prefer.

substrait/type/tests/TypeTest.cpp
chaojun-zhang: Added.

include/substrait/common/NumberUtils.h
```diff
@@ -0,0 +1,20 @@
...
+
+#pragma once
+
+#include <iostream>
+
+namespace io::substrait::common {
+
+class NumberUtils {
+ public:
+  static bool is_number(const std::string& value) {
```
westonpace, line 11: It would be more flexible to accept `std::string_view`.

include/substrait/common/NumberUtils.h
```diff
@@ -0,0 +1,20 @@
...
+    std::string::const_iterator it = value.begin();
+    while (it != value.end() && std::isdigit(*it)) {
```
westonpace, line 13: `std::isdigit` can lead to undefined behavior with `char` (`std::string::const_iterator` will give you `char`).  Given that this is likely to be called on untrusted (e.g. yaml / proto plans) input we should avoid it.  Can we cast to unsigned char or come up with some other way to do this?  Maybe use [strtol](https://en.cppreference.com/w/c/string/byte/strtol)?

This will not support decimal.  I'm not sure if that is needed.  If it isn't can we name this `is_integer`?

This will not support negative numbers.  Same as above, if that is not needed let's name this `is_nonnegative_integer`

include/substrait/common/NumberUtils.h
westonpace, line 11: Why `snake_case`?

include/substrait/type/Type.h
```diff
@@ -417,7 +419,7 @@ class StringLiteral : public ParameterizedTypeBase {
   }
 
   [[nodiscard]] bool isWildcard() const override {
```
westonpace, line 421: This isn't how I would interpret the phrase "wildcard" (but that is probably a discussion for a separate issue).  I think `any` is a wildcard and `T` is a placeholder.  Maybe placeholder would be a better term.  But maybe I am wrong in my interpretation too.

substrait/type/Type.cpp
```diff
@@ -25,9 +27,148 @@ size_t findNextComma(const std::string& str, size_t start) {
...
+          std::stoi(precision->value()), std::stoi(scale->value()), nullable);
+    } else {
+      SUBSTRAIT_FAIL(
+          "Fail decode to Decimal type, precision or scale parameter must be a positive number")
+    }
+  }
+}
+
+template <class ParameterizedTypeTag, class TypeTag, TypeKind kind>
+ParameterizedTypePtr decodeLenghBaseType(
```
westonpace, line 117: `decodeLengthBaseType`

westonpace (changes requested):
Thanks for adding the tests.  I have a few suggestions for `is_number`.

include/substrait/common/NumberUtils.h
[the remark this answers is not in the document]
chaojun-zhang: Moved

include/substrait/common/NumberUtils.h
```diff
@@ -0,0 +1,20 @@
+/* SPDX-License-Identifier: Apache-2.0 */
+
+#pragma once
+
+#include <iostream>
```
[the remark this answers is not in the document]
chaojun-zhang: Updated

include/substrait/common/NumberUtils.h
chaojun-zhang: Fixed

include/substrait/common/NumberUtils.h
chaojun-zhang: Updated

include/substrait/common/NumberUtils.h
chaojun-zhang: Updated

include/substrait/type/Type.h
chaojun-zhang: I added method  `isPlaceholder` to solve this.

substrait/type/Type.cpp
chaojun-zhang: Fixed

CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ find_package(Protobuf REQUIRED)
 include_directories(${PROTOBUF_INCLUDE_DIRS})
 
 add_subdirectory(third_party)
+include_directories(.)
```
chaojun-zhang, line 23: `include_directories(.)` seems too ambitious, may be add 'src' folder under root director and move `substrait` folder inside `src`, then replace `include_directories(.)` with `include_directories(src)`?

Anyway let's plan in another separate PR.

chaojun-zhang:
@westonpace Please help to review again, thanks

substrait/common/StringUtils.h
```diff
@@ -0,0 +1,33 @@
+/* SPDX-License-Identifier: Apache-2.0 */
+
+#pragma once
+
+#include <algorithm>
+#include <iostream>
```
westonpace, line 6:
```suggestion
```

substrait/common/NumberUtils.h
```diff
@@ -0,0 +1,24 @@
+/* SPDX-License-Identifier: Apache-2.0 */
+
+#pragma once
+
+#include <string>
```
westonpace, line 5:
```suggestion
#include <algorithm>
#include <string>
```

for std::all_of

substrait/common/NumberUtils.h
```diff
@@ -0,0 +1,24 @@
...
+namespace io::substrait::common {
+
+class NumberUtils {
+ public:
+  static bool isInteger(std::string_view s) {
+    return isNonNegativeInteger(s) || isNegativeInteger(s);
+  }
+
+  static bool isNonNegativeInteger(std::string_view s) {
+    return !s.empty() && std::all_of(s.begin(), s.end(), ::isdigit);
```
westonpace, line 16:
```suggestion
    return !s.empty() && std::all_of(s.begin(), s.end(), [](unsigned char d) {
      return std::isdigit(d);
    });
```
This is needed according to the note in https://en.cppreference.com/w/cpp/string/byte/isdigit

substrait/common/StringUtils.h
```diff
@@ -0,0 +1,33 @@
...
+    return static_cast<std::string>(res);
+  }
+
+  static std::string rtrim(std::string_view s) {
+    size_t end = s.find_last_not_of(WHITESPACE);
+    std::string_view res = (end == std::string::npos) ? "" : s.substr(0, end + 1);
+    return static_cast<std::string>(res);
+  }
+
+  static std::string trim(std::string_view s) {
```
westonpace, line 24: These methods should return `std::string_view`.  That way, you leave it up to the caller whether you materialize or not.  Also, you can then do `rtrim(ltrim(s))` without making a redundant temporary copy.

substrait/common/StringUtils.h
```diff
@@ -0,0 +1,33 @@
...
+    return rtrim(ltrim(s));
+  }
+
+  static const std::string WHITESPACE;
```
westonpace, line 28:
```suggestion
  constexpr std::string_view kWhitespace = " \n\r\t\f\v";
```

First, using `string_view` over `string` allows us to make it constexpr and there is no need to separate definition from declaration.

Second, we should be consistent and either use kConstant everywhere or CONSTANT everywhere and we are using kConstant elsewhere (e.g. Exceptions.h)

westonpace (changes requested):
A few more suggestions.  Sorry for the trouble.

westonpace (approved):
Thank you for your persistence and for this addition!

PR #22, opened 2023-02-18
refactor: move source code into a src folder

chaojun-zhang:
move source code into a src folder


CMakeLists.txt
```diff
@@ -30,4 +30,4 @@ if(${SUBSTRAIT_CPP_BUILD_TESTING})
   enable_testing()
 endif()
 
-add_subdirectory(substrait)
+add_subdirectory(src/substrait)
```
jbalint, line 33: is there any value in keeping subtrait/ under src/? why not just rename substrait to src?

EpsilonPrime: I'm currently importing substrait internal headers with:

`#include "substrait/textplan/parser/ParseResult.h"`

If we only do a rename the fact that a substrait file is being included goes away.  Also helps to mirror the path hierarchy (although we have an extra ::io in this module).

chaojun-zhang: Agreed, the key is we can refer header files in a  consistent way that always start with `substrait` prefix, whether it is either under src/substrait or under include/substrait directory.

jbalint: Makes sense. Thanks for the explanation

EpsilonPrime (approved):
Looks good to me!

westonpace (approved):
Thanks!  LGTM

PR #24, opened 2023-02-28
Add ctest option --output-on-failure

EpsilonPrime:
This should make it easier to determine the cause of CI failures.


westonpace (approved):
Yes, this will be a great help :)

PR #26, opened 2023-03-02
GH-10: Add second batch of functionality to binary to text plan conversion

EpsilonPrime:
Building on PR #23 this pull request:

- Adds support for relations, schemas, and sources.
- Does not add support for pipelines and expressions.
- Changes locations for binary to text conversion to use protobuffer message pointers.

The next PR will add the pipelines and expression capability.


chaojun-zhang:
Can we squash commits into one commit  in order to make git log history more clean?

EpsilonPrime:
I agree this is rather messy as it also includes every change to #23.  Once 23 is committed I'm hoping that 99% of the commits will just disappear.  I'm not aware of any method to make a PR dependent on another which could be useful here.  I suggest ignoring this PR until #23 is committed.

EpsilonPrime:
Closing this PR so I can create a cleaner version for easier review.

Issue #27, opened 2023-03-02
Substrait exceptions should note the code location of the caller


EpsilonPrime:
The current implementation of macros such as SUBSTRAIT_FAIL is to return an exception with the location of the exceptions library:

```
libc++abi: terminating with uncaught exception of type io::substrait::common::SubstraitRuntimeError: Exception: SubstraitRuntimeError
Error Code: INVALID_STATE
Type: system
Reason: Return type from visitExpression with case of 2 was not std::string!
Function: SubstraitException
File: /home/user/projects/substrait-cpp/src/substrait/common/Exceptions.cpp
:Line: 22

```

chaojun-zhang:
I will take and fix

EpsilonPrime:
Thanks!

PR #29, opened 2023-03-03
 Substrait exceptions should note the code location of caller

chaojun-zhang:
Substrait exceptions should note the code location of caller , see https://github.com/substrait-io/substrait-cpp/issues/27


EpsilonPrime (approved):
Thanks for fixing this so quickly!

westonpace (approved):
Thanks for adding this!

Issue #15, opened 2023-02-08
Add lint to CI


westonpace:
Required:

cmake_format
clang-format
Some kind of spdx licensing header check

Consider:

clang-tidy
iwyu

chaojun-zhang:
Is there any preset github action for this that we can refer to?

PR #30, opened 2023-03-03
Add lint to CI include checking for cmake-format, clang-format and license header

chaojun-zhang:
Add lint to CI include checking for cmake-format, clang-format and license header, see https://github.com/substrait-io/substrait-cpp/issues/15

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -1,3 +1,4 @@
+# SPDX-License-Identifier: Apache-2.0
 name: Ubuntu Build & Test
 
 on:
@@ -7,6 +8,22 @@ on:
     branches: [main]
 
 jobs:
+  check:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
+      - name: Setup Ubuntu
+        run: ./scripts/setup-ubuntu.sh
+      - name: Check License Header
+        uses: apache/skywalking-eyes/header@v0.4.0
+      - name: Check CMake files
+        run: find . \( -name '*.cmake' -o -name 'CMakeLists.txt' \) -exec cmake-format $* {} +
+      - name: Clang-tidy
+        run: python3 scripts/run-clang-tidy.py "." "build" "third_party,scripts,docker,cmake_modules" "h,hpp,cc,cpp"
+      - run: mkdir build
   build:
     runs-on: ubuntu-latest
 
--- a/.licenserc.yaml
+++ b/.licenserc.yaml
@@ -1,13 +1,22 @@
 header:
   license:
     spdx-id: Apache-2.0
-    copyright-owner: Apache Software Foundation
+    content: |
+      SPDX-License-Identifier: Apache-2.0
 
   paths-ignore:
-    - 'dist'
-    - 'licenses'
+    - '.github'
+    - '.gitignore'
+    - '.gitmodules'
+    - '.clang-format'
+    - '.clang-tidy'
+    - '.licenserc.yaml'
+    - 'third_party/fmt'
+    - 'third_party/googletest'
+    - 'third_party/substrait'
+    - 'third_party/yaml-cpp'
     - '**/*.md'
-    - 'LICENSE'
-    - 'NOTICE'
+    - '**/*.json'
+    - '**/*.log'
 
-  comment: on-failure
+  comment: never
--- a/scripts/setup-ubuntu.sh
+++ b/scripts/setup-ubuntu.sh
@@ -25,3 +25,5 @@ sudo --preserve-env apt install -y \
   libprotobuf-dev \
   libprotobuf23 \
   protobuf-compiler
+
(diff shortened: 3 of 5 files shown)


.github/workflows/build_test.yml
```diff
@@ -7,6 +7,30 @@ on:
...
+        with:
+          header:
+            license:
+              spdx-id: Apache-2.0
+      - name: Check CMake files
+        id: cmake-format
+        uses: PuneetMatharu/cmake-format-lint-action@v1.0.0
+        with:
+          args: --config-files .cmake-format.json --in-place
+      - name: Review with Clan-tidy
```
EpsilonPrime, line 25: Clang

.github/workflows/build_test.yml
```diff
@@ -7,6 +7,30 @@ on:
     branches: [main]
 
 jobs:
+  check:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check License Header
+        uses: apache/skywalking-eyes/header@main
```
westonpace, line 15: Cool find :)

Should we pin to a revision?  The example on their main page says:

```
  uses: apache/skywalking-eyes/header@main      # always prefer to use a revision instead of `main`.
```

.github/workflows/build_test.yml
```diff
@@ -7,6 +7,30 @@ on:
...
+        with:
+          header:
+            license:
+              spdx-id: Apache-2.0
+      - name: Check CMake files
+        id: cmake-format
+        uses: PuneetMatharu/cmake-format-lint-action@v1.0.0
```
westonpace, line 22: I am hesitant to rely on a 3rd party project this small.  The actual command is `find . \( -name '*.cmake' -o -name 'CMakeLists.txt' \) -exec cmake-format $* {} +`.  Can we maybe use that directly?

.github/workflows/build_test.yml
```diff
@@ -7,6 +7,30 @@ on:
...
+        with:
+          args: --config-files .cmake-format.json --in-place
+      - name: Review with Clan-tidy
+        uses: ZedThree/clang-tidy-review@v0.10.0
```
westonpace, line 26: This project appears to create a pull request review instead of a failing check?  Though maybe I am misunderstanding their github.  It's also a somewhat small project, though larger than the cmake-format one.

Nevermind, I see it fails later on if there were any comments.

This feels a little over complex.  Maybe we can get away with `run` and a call to `run-clang-tidy`?

westonpace (changes requested):
Thanks for taking this on.  Some initial thoughts.

.github/workflows/build_test.yml
chaojun-zhang: Updated

.github/workflows/build_test.yml
chaojun-zhang: Updated

.github/workflows/build_test.yml
chaojun-zhang: Fixed

chaojun-zhang:
Hi @westonpace , please help to re-run and trigger the check again.

scripts/run-clang-tidy.py
```diff
@@ -0,0 +1,59 @@
...
+print("Arguments: " + str(sys.argv))
+# Get absolute source dir after removing leading and trailing seperators from input.
+sourcedir = currentdir + sys.argv[1].lstrip(os.sep).rstrip(os.sep)
+print("Source directory: " + sourcedir)
+builddir = sourcedir + os.sep + sys.argv[2].rstrip(os.sep)
+print("Build directory: " + builddir)
+# If exclude dirs is not empty, split it into a tuple
+excludedirs = ()
+if sys.argv[3]:
+    excludedirs = tuple([(sourcedir + os.sep + s).rstrip(os.sep) for s in sys.argv[3].split(',')])
```
westonpace, line 25:
```suggestion
excludedirs = tuple([(sourcedir + os.sep + s).rstrip(os.sep) for s in sys.argv[3].split(',')])
```

Probably no need for this if check since you are relying on sys.argv[4] being present below.

westonpace (commented):
Some small suggestions and then I think this is good to go.

mbrobbel:
> I'm not sure why the build is not running. Perhaps you need to rebase to get the latest github workflows?

https://github.com/substrait-io/substrait-cpp/actions/runs/4459109898/workflow

chaojun-zhang:
https://github.com/substrait-io/substrait-cpp/pull/37 should be merge first.

westonpace:
I've merged #37.  Do you want to rebase now?  If it's still failing or we need changes to those files that are checked in then let me know and I should have some time to work it out on Friday too.

chaojun-zhang:
I've rebased, seems the build is not running now

chaojun-zhang:
Rebased and finally passed

EpsilonPrime:
I'm looking forward to this one, it'll be nice to have something to catch when I forget the license headers!

PR #31, opened 2023-03-04
Add Decimal class to initially perform proto to string conversion.

EpsilonPrime:
This class requires discussion.

Do we need a generic class for manipulating the decimal datatype or do we already have one in the ecosystem?  My current needs are to convert from the substrait proto (::substrait::proto::Expression_Literal_Decimal) into a string form (and eventually back).  Would folks need to be able to perform actual operations with this class (basic arithmetic would be likely).

Once we figure out what we want this class to do (I expect massive changes) I'll add tests.

### DIFF
--- a/include/substrait/common/Decimal.h
+++ b/include/substrait/common/Decimal.h
@@ -1,0 +1,30 @@
+/* SPDX-License-Identifier: Apache-2.0 */
+
+#pragma once
+
+#include "substrait/proto/algebra.pb.h"
+
+namespace io::substrait::common {
+
+class Decimal {
+  Decimal(std::string v, int32_t p, int32_t s)
+      : value_(std::move(v)), precision_(p), scale_(s) {}
+
+ public:
+  static Decimal fromProto(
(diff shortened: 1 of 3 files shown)


westonpace:
Hmm, ideally we wouldn't want to reinvent this ourselves.  The arrow-cpp package has utilities for decimals.  So we could use that or vendor our own.

I think the most common case where we'd want decimal manipulation (outside of actually consuming a plan) would be if we were doing any kind of expression optimization.  For example, simplifying the expression `add(add(x, 1), 2)` to `add(x, 3)` would require the ability to add `1` and `2` and get `3`.

At the moment, I think we can safely assume we will not need it (i.e. I don't think we need this for text format parsing or serialization do we?).  However, I can't rule out that we might need it someday.

I do think we want to pretty print decimal literals though so maybe that does imply some ability to manipulate decimal values.

EpsilonPrime:
Thanks, I'll close this and try to implement something closer to what Arrow does.

PR #33, opened 2023-03-09
feat: document the format of literals in the Substrait Text Plan Language

EpsilonPrime:


src/substrait/textplan/README.md
```diff
@@ -0,0 +1,208 @@
+# Substrait Text Plan Language Reference
+
+## Types
+Types are the core at the core of Substrait plans.  Functions consume data of a particular type and output data in the 
+same or a different type.  Optional types in Substrait are represented by appending a question mark to the type.
```
westonpace, line 5:
```suggestion
Types are at the core of Substrait plans.  Functions consume data of a particular type and output data in the 
same or a different type.  Nullable types in Substrait are represented by appending a question mark to the type.
```

Small grammar change in the first sentence.  More importantly, can we replace "optional" with "nullable"?  I think that is more consistent with docs elsewhere.

src/substrait/textplan/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+It's worth noting that there are no unsigned versions of numeric types in Substrait but they can be approximated with
+casts and user-defined types.
```
westonpace, line 8:
```suggestion
It's worth noting that there are no unsigned versions of numeric types in Substrait but they can be implemented with
 user-defined types.
```
No need to cast I think.  You can read directly into a user defined type.

src/substrait/textplan/README.md
```diff
@@ -0,0 +1,208 @@
...
+| timestamp            | timestamp without a timezone                                             |
+| date                 | date (day resolution)                                                    |
+| time                 | time within a day (microsecond resolution)                               |
+|                      |                                                                          |
+| interval_year        | duration specified in years and months                                   |
+| interval_day         | duration specified in days, hours, minutes, seconds, and microseconds    |
+| timestamp_tz         | timestamp with a specified timezone                                      |
+| uuid                 | 128-bit UUID                                                             |
+|                      |                                                                          |
+| varchar\<L>          | a variable sized character sequence of up to length L                    |
```
westonpace, line 33: Why `\<`? Is this a markdown escape of some kind?

src/substrait/textplan/README.md
```diff
@@ -0,0 +1,208 @@
...
+```
+123_i8
+0_i16
+123_i32
+1234_i64
+```
+
+### Floating Point Literals
+
+Floating point literals are represented with by a decimal number followed by an optional exponent.
```
westonpace, line 76: Does the exponent always need a sign char?  I recently found out that 10e7 is not a numeric YAML literal but 10e+7 is :)

src/substrait/textplan/README.md
```diff
@@ -0,0 +1,208 @@
...
+digits to translate the result by). 
+
+#### Example Decimal Literals
+
+```
+-123_decimal<3,0>
+42_decimal<5,-4>
+```
+
+### String Literals
```
westonpace, line 99: How are text plans encoded?  Can we specify UTF-8 encoding similar to JSON?

src/substrait/textplan/README.md
```diff
@@ -0,0 +1,208 @@
...
+| Escape Sequence | Description                         |
+|-----------------|-------------------------------------|
+| \n              | linefeed (ASCII code 10)            |
+| \r              | carriage return (ASCII code 13)     |
+| \t              | tab (ASCII code 8)                  |
+| \\              | backslash                           |
+| \'              | single quote                        |
+| \"              | double quote                        |
+| \xNN            | two byte hexadecimal character code |
+| \u{NNNNNNN}     | unicode code sequence               |
```
westonpace, line 113: Is this minimal set of escape sequences based on an existing implementation?  E.g. I think this looks a bit like Java.  Do we have confidence it is sufficiently complete?

src/substrait/textplan/README.md
```diff
@@ -0,0 +1,208 @@
...
+Binary literals are represented as strings containing hexadecimal digits (no escapes required).  Fixed binary literals
+are measured in bytes and thus have two hexadecimal characters per required byte.  A UUID is effectively a
+fixedbinary\<16> that ignores any provided dashes in its representation. 
+
+#### Example Binary Literals
+
+```
+"0123456789abcde"_binary
+"1234"_fixedbinary<2>
+"00-ABCD-1234"_uuid
```
westonpace, line 135: This seems too small to be a UUID?  Isn't it only 5 bytes?

src/substrait/textplan/README.md
```diff
@@ -0,0 +1,208 @@
...
+```
+
+### Time Literals
+
+There are two kinds of time literals:  literals that represent particular points in time and literals that represent
+durations.  The timestamps represent dates from the beginning of the year 1000 to the end of the year 9999.  The 
+interval_year duration has its precision in months and the interval_day duration has its precision in microseconds.
+
+To make the structured data easier to read the pseudo types ```year```, ```month```, ```day```, ```second```, and
+```microsecond``` (plural or singular) are suggested for readability.
```
westonpace, line 145: Suggested?  Or required?

src/substrait/textplan/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+#### Example Time Literals
+
+```
+"1000-01-01 00:00:00.000000"_timestamp
+"1000-01-01 00:00:00.000000 UTC"_timestamp_tz
```
westonpace, line 151: This looks very close to ISO-8601.  Can we just strictly require ISO-8601?

src/substrait/textplan/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+### Structured Literals
+
+Structure literals use the brace syntax.  Types are not necessary within the value section as the types should be 
+homogeneous if the structure is well-formed. 
+
+|                                                           | Description                                   |
+|-----------------------------------------------------------|-----------------------------------------------|
+| {}_list\<string>                                          | empty list of string                          |
+| {}_list\<string?>                                         | empty list of optional strings                |
```
westonpace, line 169:
```suggestion
| {}_list\<string?>                                         | empty list of nullable strings                |
```

src/substrait/textplan/README.md
```diff
@@ -0,0 +1,208 @@
...
+| {null, "a", "b"}_list\<string?>                           | list of optional strings                      |
```
westonpace, line 170:
```suggestion
| {null, "a", "b"}_list\<string?>                           | list of nullable strings                      |
```

src/substrait/textplan/README.md
```diff
@@ -0,0 +1,208 @@
...
+| {}_map\<int, string>                                      | empty map with integer keys and string values |
+| {42->"life", 32->"everything"}_map\<int, string>          | map with integer keys and string values       |
```
westonpace, line 172: Hmm...why `->` instead of `:` which is probably more universally expected?

src/substrait/textplan/README.md
```diff
@@ -0,0 +1,208 @@
...
+| {"a", {"b", "c"}_struct\<string, struct\<string, string>> | structure with nested substructure            |
```
westonpace, line 173:
```suggestion
| {"a", {"b", "c"}}_struct\<string, struct\<string, string>> | structure with nested substructure            |
```

src/substrait/textplan/README.md
```diff
@@ -0,0 +1,208 @@
...
+All null types must include an optional value (so that the null can be provided) and thus normally would end with a 
+question mark to show that it is optional.  For brevity the question mark may be eliminated.
+
+Nulls can represent either an unknown value or the lack of a value depending on the semantics of the operation.  (A sum
+function using null as unknown would result in unknown, but a sum operation treating null as not present would result in
+the total of the non-null elements.)
+
+#### Example Null Literals
+```
+null_i32
```
westonpace, line 205: Can you add a comment mentioning that the type is implicitly `i32?`?

src/substrait/textplan/README.md
```diff
@@ -0,0 +1,208 @@
...
+true
+false
+true_bool
+false_bool
+```
+
+### Null
+
+Null also has a distinct form.  Types are still necessary to inform which function is to be used to process the literal.
+All null types must include an optional value (so that the null can be provided) and thus normally would end with a
```
westonpace, line 196:
```suggestion
All null types must be nullable (so that the null can be provided) and thus normally would end with a 
```

westonpace (commented):
Thanks!  Sorry for the delay in reviewing.

src/substrait/textplan/README.md
```diff
@@ -0,0 +1,208 @@
...
+Types are important in Substrait plans as the choice of type controls which implementation of a given function should
+be used.  In the text representation of plans we label literals with their types to make it clear which function should
+be invoked within the plan. The type of a literal is appended to the literal behind an underscore (_).  Typed literals 
+may not be optional.
+
+### Integer Literals
+
+There are four different integer widths.  While other languages provide for binary, octal, and hexadecimal
+representations of their integer literals (as well as spacing separators) human readability of literals isn't a current
+requirement of the language and so are omitted for brevity.
```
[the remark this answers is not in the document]
EpsilonPrime: Now specifically calls out that the alternative representations that other languages have (binary, octal, and hexidecimal) are not supported.

src/substrait/textplan/README.md
EpsilonPrime: Ah, that's to keep the hexadecimal literals distinct from the floating point ones.  It's not an issue if we don't have hexadecimal literals but might as well make the distinction in case we decide to upgrade later on.

src/substrait/textplan/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+Decimal literals are similar to that of floating point literals however are represented by a real number with an
+exponent which is specified as part of the type.  The type includes the precision and the scale (number of base 10
+digits to translate the result by). 
+
+#### Example Decimal Literals
+
+```
+-123_decimal<3,0>
+42_decimal<5,-4>
```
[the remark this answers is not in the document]
EpsilonPrime: Done.  Also moved the parameters next to the value instead of treating them as a decimal subtype.

src/substrait/textplan/README.md
```diff
@@ -0,0 +1,208 @@
...
+```
+
+### String Literals
+
+No need for multiline.  Raw strings may be useful for regexes.  Strings with no size limit may be specified without a
```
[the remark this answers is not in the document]
EpsilonPrime: I've defined raw literals based on the Java multiline style.

src/substrait/textplan/README.md
EpsilonPrime: Similar to structured literals in the following section types aren't needed because the type identifier already explains what is inside.  Effectively an interval_year or interval_day is merely a well-known structure.  If one only wanted to provide part of the structure the types would again be necessary (although now that I'm thinking about it we're actually providing names for this purpose here).

src/substrait/textplan/README.md
EpsilonPrime: done.

src/substrait/textplan/README.md
EpsilonPrime: changed

src/substrait/textplan/README.md
EpsilonPrime: done

src/substrait/textplan/README.md
```diff
@@ -0,0 +1,208 @@
...
+```
+
+### Structured Literals
+
+Structure literals use the brace syntax.  Types are not necessary within the value section as the types should be 
+homogeneous if the structure is well-formed. 
+
+|                                                           | Description                                   |
+|-----------------------------------------------------------|-----------------------------------------------|
+| {}_list\<string>                                          | empty list of string                          |
```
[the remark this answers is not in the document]
EpsilonPrime: Right now I've defined all of the datastructures in the same way -- lists, maps, structs are all defined with curly braces (somewhat like the way you can brace initialize in C++).  This decision can be altered if needed.

src/substrait/textplan/README.md
```diff
@@ -0,0 +1,208 @@
...
+false
+true_bool
+false_bool
+```
+
+### Null
+
+Null also has a distinct form.  Types are still necessary to inform which function is to be used to process the literal.
+All null types must include an optional value (so that the null can be provided) and thus normally would end with a 
+question mark to show that it is optional.  For brevity the question mark may be eliminated.
```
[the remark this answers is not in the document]
EpsilonPrime: Completely reworded.

src/substrait/textplan/README.md
EpsilonPrime: done

src/substrait/textplan/README.md
EpsilonPrime: done as part of the rewording

src/substrait/textplan/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+Note that user defined types are not yet supported in text plans (but will be at a future date).
+
+#### Example Types
+
+|                                        | Explanation                                                         |
+|----------------------------------------|---------------------------------------------------------------------|
+| i64\?                                  | optional int64                                                      |
+| varchar\<5>                            | a string with up to five characters                                 |
+| struct\<string,i8,i32?,timestamp_tz>   | a structure with a string, two ints, one optional, and a timestamp  |
```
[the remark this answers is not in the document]
EpsilonPrime: done

src/substrait/textplan/README.md
```diff
@@ -0,0 +1,208 @@
...
+| struct\<string,struct\<string,string>> | a structure containing a string and a substructure with two strings |
+
+## Literals
+
+Types are important in Substrait plans as the choice of type controls which implementation of a given function should
+be used.  In the text representation of plans we label literals with their types to make it clear which function should
+be invoked within the plan. The type of a literal is appended to the literal behind an underscore (_).  Typed literals 
+may not be optional.
```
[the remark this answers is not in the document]
EpsilonPrime: removed this line

src/substrait/textplan/README.md
westonpace: Ok.  It appears `\` escapes are valid in common-mark (even though inline HTML is not) so this should be ok.

EpsilonPrime:
Decimals have been updated again as discussed.

src/substrait/textplan/README.md
```diff
@@ -0,0 +1,216 @@
...
+
+```
+-123_decimal<3,0>  // -123
+42_decimal<5,-4>  // 420000
+```
+
+### String Literals
+
+Strings with no size limit may be specified without a type specifier for brevity.  Substrait text plans are UTF-8
+encoded throughout so any characters contained in strings must also co valid UTF-8.  For byte content that is not in
```
westonpace, line 103:
```suggestion
encoded throughout so any characters contained in strings must also be valid UTF-8.  For byte content that is not in
```

westonpace (approved):

PR #34, opened 2023-03-16
feat: functions and literals in text format

EpsilonPrime:
* Adds support for functions
* Adds support for most literal types (including Decimal)
* Quality improvements
   * Adds a new matcher that allows only symbols of a particular type to be matched
   * Added PROTO_LOCATION macro to make specifying locations more convenient.
   * Upgraded the converter tool to accept globs (only useful for testing as shells already expand globs).
   * Many error messages are now returned to the ParseResult instead of being left in text for a human to detect.

### DIFF
--- a/src/substrait/expression/DecimalLiteral.cpp
+++ b/src/substrait/expression/DecimalLiteral.cpp
@@ -38,7 +38,6 @@ bool DecimalLiteral::isValid() {
   return value_.size() == 16 && precision_ >= 1 && precision_ <= 38;
 }
 
-
 std::string DecimalLiteral::toBaseString() {
   std::stringstream decimalString;
   if (value_.empty()) {
@@ -56,7 +55,7 @@ std::string DecimalLiteral::toBaseString() {
   }
 
   // Collect the bytes into an unsigned integer.
-  absl::uint128 value = 0;  // Will only hold 16 hex chars.
+  absl::uint128 value = 0; // Will only hold 16 hex chars.
   for (size_t i = numBytes; i > 0; i--) {
     value = (value << 8) | static_cast<unsigned char>(processingValue[i - 1]);
   }
@@ -93,4 +92,4 @@ std::string DecimalLiteral::toString() {
   return decimalString.str();
 }
 
-} // namespace io::substrait::common
+} // namespace io::substrait::expression
--- a/src/substrait/expression/DecimalLiteral.cpp
+++ b/src/substrait/expression/DecimalLiteral.cpp
@@ -11,8 +11,8 @@ namespace io::substrait::expression {
 
 namespace {
 
-// invertBits flips the sign of a two-complements value.
-std::string invertBits(const std::string& value) {
+// negate flips the sign of a two-complements value.
+std::string negate(const std::string& value) {
   std::string newValue = value;
   // Flip all of the bits and add one.
   bool carryover = true;
@@ -51,7 +51,7 @@ std::string DecimalLiteral::toBaseString() {
   bool isNegative = ((value_[numBytes - 1] & 0x80) != 0);
   if (isNegative) {
     decimalString << "-";
-    processingValue = invertBits(value_);
+    processingValue = negate(value_);
   }
 
   // Collect the bytes into an unsigned integer.
--- a/src/substrait/expression/DecimalLiteral.cpp
+++ b/src/substrait/expression/DecimalLiteral.cpp
@@ -1,0 +1,96 @@
+/* SPDX-License-Identifier: Apache-2.0 */
+
+#include "substrait/expression/DecimalLiteral.h"
+
+#include <sstream>
+
+#include "absl/numeric/int128.h"
+#include "substrait/proto/algebra.pb.h"
+
+namespace io::substrait::expression {
+
+namespace {
+
+// invertBits flips the sign of a two-complements value.
+std::string invertBits(const std::string& value) {
+  std::string newValue = value;
+  // Flip all of the bits and add one.
+  bool carryover = true;
+  for (char& b : newValue) {
+    uint8_t newB = ~(static_cast<uint8_t>(b));
+    if (carryover) {
+      newB++;
+    }
+    carryover = carryover && (newB == 0);
+    b = static_cast<char>(newB & 0xff);
+  }
+  return newValue;
+}
+
+} // namespace
+
+DecimalLiteral DecimalLiteral::fromProto(
+    const ::substrait::proto::Expression_Literal_Decimal& proto) {
+  return {proto.value(), proto.precision(), proto.scale()};
+}
+
+bool DecimalLiteral::isValid() {
+  return value_.size() == 16 && precision_ >= 1 && precision_ <= 38;
+}
+
+
+std::string DecimalLiteral::toBaseString() {
+  std::stringstream decimalString;
+  if (value_.empty()) {
+    return "0";
+  }
+
+  std::string processingValue = value_;
+
+  // Determine the sign of the value.
+  size_t numBytes = value_.size();
+  bool isNegative = ((value_[numBytes - 1] & 0x80) != 0);
+  if (isNegative) {
+    decimalString << "-";
+    processingValue = invertBits(value_);
+  }
+
+  // Collect the bytes into an unsigned integer.
+  absl::uint128 value = 0;  // Will only hold 16 hex chars.
+  for (size_t i = numBytes; i > 0; i--) {
+    value = (value << 8) | static_cast<unsigned char>(processingValue[i - 1]);
+  }
+
+  // Pull off the digits backwards.
+  std::stringstream valueString;
+  while (value >= 10) {
+    valueString << static_cast<char>('0' + (value % 10));
+    value /= 10;
+  }
+  valueString << static_cast<char>('0' + value);
+
+  // Reverse the digits.
+  std::string v = valueString.str();
+  for (auto c = v.rbegin(); c != v.rend(); c++) {
+    decimalString << *c;
+  }
+
+  return decimalString.str();
+}
+
+std::string DecimalLiteral::toString() {
+  std::stringstream decimalString;
+
+  decimalString << toBaseString();
+
+  if (scale_ > 0) {
+    decimalString << "E-" << scale_;
+  } else if (scale_ < 0) {
+    decimalString << "E+" << -scale_;
+  }
+  decimalString << "@precision=" << precision_;
+
+  return decimalString.str();
+}
+
+} // namespace io::substrait::common
--- a/src/substrait/textplan/Location.cpp
+++ b/src/substrait/textplan/Location.cpp
@@ -6,7 +6,7 @@
 
 namespace io::substrait::textplan {
 
-constexpr const Location Location::kUnknownLocation(
+constexpr Location Location::kUnknownLocation(
     static_cast<google::protobuf::Message*>(nullptr));
 
 bool operator==(const Location& c1, const Location& c2) {
--- a/src/substrait/textplan/Location.cpp
+++ b/src/substrait/textplan/Location.cpp
@@ -6,7 +6,7 @@
 
 namespace io::substrait::textplan {
 
-const Location Location::kUnknownLocation(
+constexpr const Location Location::kUnknownLocation(
     static_cast<google::protobuf::Message*>(nullptr));
 
 bool operator==(const Location& c1, const Location& c2) {
--- a/src/substrait/textplan/converter/PlanPrinterVisitor.cpp
+++ b/src/substrait/textplan/converter/PlanPrinterVisitor.cpp
@@ -3,33 +3,51 @@
 #include "substrait/textplan/converter/PlanPrinterVisitor.h"
 
 #include <iterator>
-#include <optional>
 #include <sstream>
 #include <string>
 
+#include "substrait/expression/DecimalLiteral.h"
 #include "substrait/proto/ProtoUtils.h"
 #include "substrait/proto/algebra.pb.h"
 #include "substrait/textplan/Any.h"
 
 namespace io::substrait::textplan {
 
+namespace {
+
+std::string stringEscape(std::string_view str) {
+  // TODO - Implement all of the escapes defined in the document.
+  std::stringstream result;
+  for (auto c : str) {
+    if (c == '"') {
+      result << '\\';
+    }
+    result << c;
+  }
+  return result.str();
+}
+
+} // namespace
+
 std::string PlanPrinterVisitor::printRelation(
-    const std::string& symbolName,
     const ::substrait::proto::Rel* relation) {
   std::stringstream text;
-
+  auto symbol = symbolTable_->lookupSymbolByLocation(PROTO_LOCATION(*relation));
   text << ::substrait::proto::relTypeCaseName(relation->rel_type_case())
-       << " relation " << symbolName << " {\n";
-  auto symbol = symbolTable_->lookupSymbolByLocation(
-      Location((google::protobuf::Message*)&relation));
+       << " relation " << symbol.name << " {\n";
+
+#if 0
   if (symbol != SymbolTable::kUnknownSymbol) {
     text << "  source " << symbol.name << ";\n";
   }
+#endif
   auto result = this->visitRelation(*relation);
   if (result.type() != typeid(std::string)) {
-    return "ERROR:  Relation subtype " +
+    errorListener_->addError(
+        "ERROR:  Relation subtype " +
         ::substrait::proto::relTypeCaseName((relation->rel_type_case())) +
-        " not supported!\n";
+        " not supported!");
+    return "";
   }
   text << ANY_CAST(std::string, result);
   text << "}\n";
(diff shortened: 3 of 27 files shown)


src/substrait/textplan/converter/PlanPrinterVisitor.cpp
```diff
@@ -7,58 +7,567 @@
...
+      text << "\"" << stringEscape(literal.fixed_char()) << "\"_fixedchar<"
+           << literal.fixed_char().length() << ">";
+      break;
+    case ::substrait::proto::Expression_Literal::kVarChar:
+      text << "\"" << stringEscape(literal.var_char().value()) << "\"_varchar<"
+           << literal.var_char().length() << ">";
+      break;
+    case ::substrait::proto::Expression_Literal::kDecimal: {
+      auto dec = common::Decimal::fromProto(literal.decimal());
+      return dec.toString();
```
westonpace, line 268: This doesn't match the format of decimals defined in https://github.com/substrait-io/substrait-cpp/pull/33/files#diff-94fce6a3a4a591e4cd44ad4364294611a3ad6cd117450d90d3888933b5230f7aR96

src/substrait/textplan/converter/PlanPrinterVisitor.cpp
```diff
@@ -7,58 +7,567 @@
...
+    case ::substrait::proto::Expression::FieldReference::kMaskedReference:
+      error_listener_->addError(
+          "Masked reference not yet supported: " + ref.ShortDebugString());
+      return "MASKED_REF_NOT_SUPPORTED";
+    case ::substrait::proto::Expression::FieldReference::REFERENCE_TYPE_NOT_SET:
+      error_listener_->addError("Reference type missing from field reference.");
+      return "";
+  }
+  error_listener_->addError(
+      "New field reference type not yet supported: " + ref.ShortDebugString());
```
westonpace, line 311:
```suggestion
      "Field reference type not yet supported: " + ref.ShortDebugString());
```

src/substrait/textplan/converter/PlanPrinterVisitor.cpp
```diff
@@ -7,58 +7,567 @@
...
+        text << ANY_CAST(std::string, visitExpression(arg.value()));
+        break;
+      case ::substrait::proto::FunctionArgument::ARG_TYPE_NOT_SET:
+        error_listener_->addError("Argument type missing the type field.");
+        text << "SCALAR_FUNCTION_ARGUMENT_TYPE_MISSING";
+        break;
+    }
+    hasPreviousText = true;
+  }
+  for (const auto& arg : function.args()) {
```
westonpace, line 344: One of `arg` or `arguments` is deprecated.  I forget which one.  You can probably just ignore the deprecated one.

src/substrait/textplan/converter/PlanPrinterVisitor.cpp
```diff
@@ -7,58 +7,567 @@
...
+    if (hasPreviousText)
+      text << ", ";
+    text << ANY_CAST(std::string, visitExpression(arg));
+    hasPreviousText = true;
+  }
+  // TODO -- Determine if the output type can be automatically determined.
```
westonpace, line 350: Why wouldn't we be able to print the return type?

src/substrait/textplan/converter/PlanPrinterVisitor.cpp
```diff
@@ -7,58 +7,567 @@
...
+  // text << "->" << visitType(function.output_type());
+
+  if (!hasPreviousText) {
```
westonpace, line 353: Looks like you handle options in the aggregate case.  Why not here?

src/substrait/textplan/converter/PlanPrinterVisitor.cpp
```diff
@@ -7,58 +7,567 @@
...
+  text << ")";
+  for (const auto& option : function.options()) {
+    text << "#" << option.name();
+    for (const auto& pref : option.preference()) {
+      text << ";" << pref;
+    }
+  }
+  text << "->" << ANY_CAST(std::string, visitType(function.output_type()));
+  // TODO -- Emit the requested sort behavior here.
+  text << "@" << ::substrait::proto::AggregationPhase_Name(function.phase());
```
westonpace, line 519: Bleh...I think this will be a little wordy.  Maybe some shorthand here at some point? Also, we can probably default / leave blank if it is `AGGREGATION_PHASE_INITIAL_TO_RESULT`.  Although the docs say the default is `INTERMEDIATE_TO_RESULT` :frowning_face:

src/substrait/textplan/tests/ParseResultMatchers.cpp
```diff
@@ -64,6 +65,44 @@ bool StringEqSquashingWhitespace(
...
+  std::vector<std::string> names;
+  for (const auto& symbol : symbols) {
+    if (types.find(symbol->type) != types.end()) {
+      names.push_back(symbol->name);
+    }
+  }
+  return names;
+}
+
+std::vector<std::string> OrderedSetDifference(
```
westonpace, line 80:
```suggestion
std::vector<std::string> orderedSetDifference(
```

Or, if we are trying to be compatible with matchers in general, maybe change `symbolNamesWithTypes` to `SymbolNamesWithTypes`?

westonpace, line 80: Why ordered?  Isn't this just "set difference" and the fact that it uses an ordered set is an implementation detail?

westonpace (changes requested):
This is getting to be quite a large project now :laughing:

src/substrait/common/Decimal.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+namespace io::substrait::common {
+
+Decimal Decimal::fromProto(
+    const ::substrait::proto::Expression_Literal_Decimal& proto) {
+  return {proto.value(), proto.precision(), proto.scale()};
+};
+
+std::string Decimal::toString() {
+  if (value_.empty()) {
```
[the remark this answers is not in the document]
EpsilonPrime: I added an optional validity check (optional because a system may be able to use the unchecked version for increased speed).  Regardless this simple check guards a potential access violation so it's hard for me to remove.

src/substrait/common/Decimal.cpp
```diff
@@ -0,0 +1,49 @@
...
+    return "0";
+  }
+
+  int32_t numBytes = value_.size() & 0xFF;  // Enough for our 16 bytes.
```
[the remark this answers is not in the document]
EpsilonPrime: The most compelling reason I have so far is that having the shorter value makes it easier to test the code (with the added tests).  I'm mostly protecting against someone constructing the protos incorrectly (or someone coming along and modifying a related portion of code to have new behavior).

src/substrait/common/Decimal.cpp
```diff
@@ -0,0 +1,49 @@
...
+  bool isNegative = ((value_[numBytes - 1] & 0x80) != 0);
+
+  std::string decimalString;
+
+  unsigned long long value = 0;
+  for (int32_t i = numBytes - 1; i >= 0; i--) {
+    value = (value << 8) | static_cast<unsigned char>(value_[i]);
+    while (value > 10) {
```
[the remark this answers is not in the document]
EpsilonPrime: Done.

src/substrait/common/Decimal.cpp
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  while (value > 0) {
+    decimalString += static_cast<char>('0' + (value % 10));
+    value /= 10;
+  }
+
+  if (decimalString.empty()) {
+    return "0";
+  }
```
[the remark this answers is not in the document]
EpsilonPrime: Done.

src/substrait/common/Decimal.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+  decimalString += "E" + std::to_string(scale_);
```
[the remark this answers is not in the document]
EpsilonPrime: Done.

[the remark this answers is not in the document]
EpsilonPrime: Done.

src/substrait/common/Decimal.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "substrait/common/Decimal.h"
+
+namespace io::substrait::common {
+
+Decimal Decimal::fromProto(
+    const ::substrait::proto::Expression_Literal_Decimal& proto) {
+  return {proto.value(), proto.precision(), proto.scale()};
+};
+
+std::string Decimal::toString() {
```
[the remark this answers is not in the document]
EpsilonPrime: With this implementation I'm trying to avoid any normalization and the documented form in #33 matches that.  The counterpoint is text format readability.  How about we leave the simplistic form for now and add making decimal strings more readable a separate task?

src/substrait/textplan/tests/ParseResultMatchers.cpp
EpsilonPrime: I chose ordered because the difference is displayed in the same order that the values were provided for comparison.  (As opposed to being in a random or sorted order.). This helps me see which ones are missing in the way the program should be processing them.

EpsilonPrime: Done.

src/substrait/textplan/converter/PlanPrinterVisitor.cpp
EpsilonPrime: The return type can (and was printed) at one point.  But it was verbose.  I'm wondering if it is possible to divine the type automatically which means we wouldn't need to expose it.

src/substrait/textplan/converter/PlanPrinterVisitor.cpp
EpsilonPrime: Done.

src/substrait/textplan/converter/PlanPrinterVisitor.cpp
EpsilonPrime: Yep, I'm getting warnings about that when I compile.  As discussed offline that's intended so I can read in older plans (like I found in Velox's tests) and roundtrip them into present-form as a side effect.  I'm hoping that the upgrade strategy should be for the parsing to handle older forms to help provide an upgrade path.

src/substrait/textplan/converter/PlanPrinterVisitor.cpp
```diff
@@ -7,58 +7,567 @@
...
+      if (literal.interval_day_to_second().microseconds() != 0) {
+        if (hasPreviousText)
+          text << ", ";
+        text << literal.interval_day_to_second().days() << "microseconds";
+      }
+      text << "}_interval_day"; // TODO - Change spec to better name.
+      break;
+    }
+    case ::substrait::proto::Expression_Literal::kFixedChar:
+      // TODO - Should the fixedchar length match its use instead?
```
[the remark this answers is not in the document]
EpsilonPrime: Reworded to "Consider dropping the length in the type in favor of strlen."  Basically the literal string has a length.  Perhaps that's the actual type length instead of something like "fixedchar<21>".  Which means we could just say fixedchar instead.

src/substrait/textplan/converter/PlanPrinterVisitor.cpp
```diff
@@ -7,58 +7,567 @@
...
+      if (literal.interval_day_to_second().seconds() != 0) {
+        if (hasPreviousText)
+          text << ", ";
+        text << literal.interval_day_to_second().days() << "seconds";
+        hasPreviousText = true;
+      }
+      if (literal.interval_day_to_second().microseconds() != 0) {
+        if (hasPreviousText)
+          text << ", ";
+        text << literal.interval_day_to_second().days() << "microseconds";
```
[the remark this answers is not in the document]
EpsilonPrime: Done.

src/substrait/textplan/converter/PlanPrinterVisitor.cpp
```diff
@@ -7,58 +7,567 @@
...
+      text << "{";
+      bool hasPreviousText = false;
+      if (literal.interval_day_to_second().days() != 0) {
+        text << literal.interval_day_to_second().days() << "days";
+        hasPreviousText = true;
+      }
+      if (literal.interval_day_to_second().seconds() != 0) {
+        if (hasPreviousText)
+          text << ", ";
+        text << literal.interval_day_to_second().days() << "seconds";
```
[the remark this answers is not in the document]
EpsilonPrime: Done.

src/substrait/textplan/converter/PlanPrinterVisitor.cpp
```diff
@@ -7,58 +7,567 @@
...
+      break;
+    case ::substrait::proto::Expression::Literal::kFp64:
+      text << literal.fp64() << "_fp64";
+      break;
+    case ::substrait::proto::Expression::Literal ::kDate: {
+      // TODO -- Format this as a date instead of a delta since an epoch.
+      if (literal.date() >= 0) {
+        text << "\"epoch+" << literal.date() << " days\"_date";
+      } else {
+        text << "\"epoch" << literal.date() << " days\"_date";
```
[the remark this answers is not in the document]
EpsilonPrime: If the date value is negative there should be a minus sign provided for free.

src/substrait/textplan/converter/PlanPrinterVisitor.cpp
```diff
@@ -7,58 +7,567 @@
...
+    case ::substrait::proto::Type::kUuid:
+    case ::substrait::proto::Type::kFixedBinary:
+    case ::substrait::proto::Type::kStruct:
+    case ::substrait::proto::Type::kList:
+    case ::substrait::proto::Type::kMap:
+    case ::substrait::proto::Type::kUserDefined:
+    case ::substrait::proto::Type::kUserDefinedTypeReference:
+      error_listener_->addError(
+          "Unsupported type requested: " + type.ShortDebugString());
+      return std::string("UNSUPPORTED_TYPE");
```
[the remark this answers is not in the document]
EpsilonPrime: Yes, most of these are not yet implemented (destined for the full-featured converter).  Would you prefer a different wording?

src/substrait/textplan/converter/PlanPrinterVisitor.cpp
```diff
@@ -7,58 +7,567 @@
...
+        RelTypeCaseName((relation->rel_type_case())) + " not supported!");
+    return "";
   }
   text << ANY_CAST(std::string, result);
   text << "}\n";
 
   return text.str();
 }
 
+std::string PlanPrinterVisitor::lookupFieldReference(uint32_t field_reference) {
```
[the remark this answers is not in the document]
EpsilonPrime: Yep, looks like I've merged the semantics of schemas and field references.  I will need to address this.

src/substrait/textplan/converter/PlanPrinterVisitor.cpp
EpsilonPrime: Agreed this part of the textplan is harder to consume, especially if we have to use the full enum names.  Added to the list.

src/substrait/textplan/converter/PlanPrinterVisitor.cpp
EpsilonPrime: I probably didn't encounter any options in the test plans I was working with.  It's not hard to add but it's one of the items I was planning to address for the fully featured version.

include/substrait/common/Decimal.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include <string>
+
+namespace substrait::proto {
+  class Expression_Literal_Decimal;
+}
+
+namespace io::substrait::common {
+
+class Decimal {
```
chaojun-zhang, line 13: I think this class should be placed under the expression directory, which  represent the c++ classes corresponding to the substrait expressions?

src/substrait/common/CMakeLists.txt
```diff
@@ -10,6 +10,15 @@ target_link_libraries(
         substrait_common
         fmt::fmt-header-only)
 
+add_library(
```
chaojun-zhang, line 13: we already have the substrait_type library under the substrait/type directory, like I mentioned previous, can we place the Decimal under substrait/expression directory ?

EpsilonPrime: @chaojun-zhang I definitely can see the argument for not placing it in the common directory and I missed that we had a similarly named library already so I'm ready to move them.  The substrait/expression directory does not yet exist -- would you like me to move this file to start it?

EpsilonPrime: Done.

src/substrait/textplan/converter/PlanPrinterVisitor.cpp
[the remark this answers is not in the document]
EpsilonPrime: The solution for this is going to store the names information in the related data for the relations (which is getting introduced and then modified in two currently pending PRs).  I'm going to fix this behavior later on.  The current implementation works only if there is only one join which is sufficient for what I've tested with so far.  So, yes, another TODO here.

src/substrait/common/Decimal.cpp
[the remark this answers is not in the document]
EpsilonPrime: I was rereading the SQL->Substrait documentation on the website and found that we refer to decimal types as decimal(10, 2) meaning that the precision and scale are considered to be part of the type.  If this is right approach, then what should be here would be 12345_decimal<5,2> (which was what was in the literal docs until I changed it to match here).  I'm leaning towards this although it make it more annoying to specify decimals (and handling decimal as a type will also be more complicated).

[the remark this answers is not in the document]
EpsilonPrime: I've been thinking about how decimal is used and I've decided both the approach taken here and the one in substrait is probably wrong.  The type portion of decimal is essentially the decimal type and its precision (which can be contained within particular implementations of decimal including decimal128 and decimal256 but more on these later).  The scale is just a convenience factor showing where the decimal point is and if you combine two decimals the scale does not need to remain the same.  Thus 123.45_decimal<5> is closer to what we should be using and the substrait definition needs updating as well.

decimal128 and decimal256 are physical versions of the type logical types decimal<38> and decimal<76> (and any precision up through 76 could be handled by decimal256).  I don't see a notion of decimal256 in substrait at this time so sticking to decimal would be fine for now.  However at the point decimal256 is introduced we would likely want to have the specific types as we have i8, i16, and i32 today:
`123.45_decimal128<5>
123.45_decimal256<5>`

EpsilonPrime:
The two major review items not addressed by the updates are the changes to the field references (implemented in a PR that's waiting for #33, #34, and #35 to merge first) and the format of Decimal (waiting for at least #33 but given that it's going to depend on a breaking change to substrait should be addressed separately after this repo has been updated to the latest substrait version).

So this should be ready for rereview.

src/substrait/textplan/converter/PlanPrinterVisitor.cpp
westonpace: Got it.  From the TODO comment I thought there was some trick that made `output_type()` incorrect in some situations.  However, your clarification suggests this is just disabled for brevity.

src/substrait/expression/Decimal.cpp
```diff
@@ -0,0 +1,96 @@
...
+    }
+    carryover = carryover && (newB == 0);
+    b = static_cast<char>(newB & 0xff);
+  }
+  return newValue;
+}
+
+} // namespace
+
+Decimal Decimal::fromProto(
```
chaojun-zhang, line 32: Can we rename to DecimalLiteral to keep in line with substrait proto?

EpsilonPrime: Done.

chaojun-zhang:
LGTM

chaojun-zhang (approved):

src/substrait/expression/DecimalLiteral.cpp
```diff
@@ -0,0 +1,95 @@
+/* SPDX-License-Identifier: Apache-2.0 */
+
+#include "substrait/expression/DecimalLiteral.h"
+
+#include <sstream>
+
+#include "absl/numeric/int128.h"
```
westonpace, line 7: Abseil is a pretty large dependency to pull in for int128 support.  I'd rather we just vendor `int128.h`.  However, I think we can defer that for a follow-up.  I'll try and do it myself.  Or do you expect that we will be needing many more of abseil's capabilities before we're done?

src/substrait/textplan/Location.cpp
```diff
@@ -6,7 +6,7 @@
 
 namespace io::substrait::textplan {
 
-const Location Location::kUnknownLocation(
+constexpr const Location Location::kUnknownLocation(
```
westonpace, line 9:
```suggestion
constexpr Location Location::kUnknownLocation(
```

I could be wrong but I think, in this case, `constexpr const` is redundant.

src/substrait/textplan/converter/PlanPrinterVisitor.cpp
```diff
@@ -3,63 +3,596 @@
...
-       << " relation " << symbolName << " {\n";
-  auto symbol = symbolTable_->lookupSymbolByLocation(
-      Location((google::protobuf::Message*)&relation));
+       << " relation " << symbol.name << " {\n";
+
+#if 0
   if (symbol != SymbolTable::kUnknownSymbol) {
     text << "  source " << symbol.name << ";\n";
   }
+#endif
```
westonpace, line 43:
```suggestion
```

src/substrait/textplan/converter/Tool.cpp
```diff
@@ -1,6 +1,7 @@
 /* SPDX-License-Identifier: Apache-2.0 */
 
 #include <getopt.h>
+#include <glob.h>
```
westonpace, line 4: Are either of these supported on Windows?

westonpace (approved):
A few minor comments but the overall structure is good and I'd rather move forward and tackle the details / follow-ups later so we can have smaller PRs.

src/substrait/textplan/converter/Tool.cpp
EpsilonPrime: Dropped getopt as it isn't needed yet and created a non-glob version for Windows (as compared to implementing an equivalent version with FindFirstFile).

src/substrait/textplan/converter/PlanPrinterVisitor.cpp
EpsilonPrime: Done.

src/substrait/textplan/Location.cpp
EpsilonPrime: Done.  Only needed for pointers apparently.

src/substrait/expression/DecimalLiteral.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+} // namespace
+
+DecimalLiteral DecimalLiteral::fromProto(
+    const ::substrait::proto::Expression_Literal_Decimal& proto) {
+  return {proto.value(), proto.precision(), proto.scale()};
+}
+
+bool DecimalLiteral::isValid() {
+  return value_.size() == 16 && precision_ >= 1 && precision_ <= 38;
```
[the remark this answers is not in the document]
EpsilonPrime: Similar to how we handle round, I believe scale can be negative to move the decimal place left of the decimal point (adding powers of 10).  Precision and scale don't appear to be related as you could be working with something with the scale of microseconds but only have 2 digits of precision.

src/substrait/expression/DecimalLiteral.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+#include "absl/numeric/int128.h"
+#include "substrait/proto/algebra.pb.h"
+
+namespace io::substrait::expression {
+
+namespace {
+
+// invertBits flips the sign of a two-complements value.
+std::string invertBits(const std::string& value) {
```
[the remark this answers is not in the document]
EpsilonPrime: Done.

src/substrait/expression/DecimalLiteral.cpp
EpsilonPrime: int128, status, and time are the most interesting to me but only int128 is needed so separating it out is fine.  Strings is powerful but we have most of that in modern C++ compilers.

src/substrait/expression/DecimalLiteral.cpp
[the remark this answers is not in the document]
westonpace: I see what you're saying but the spec does say 

image

and this is in line with other engines such as SQL server:

image

I think the basic problem is that `P` is the "number of digits" and not "number of significant digits".  So you can't have decimal<2, -2> to represent something like `1100` (eleven hundred) because then you'd have 4 digits.

However, we can worry about this in follow-up / future PRs.  I'll open an issue.

PR #38, opened 2023-03-26
feat(lint): Enable Clang tidy check and Cmake check and clang format

chaojun-zhang:
This PR fixes the problems found by clang-tidy, and it's open for an in-depth discussion about the  check list defined in `.clang-tidy`.

This PR also format cmake files  with cmake-format and format code with clang-format


include/substrait/type/Type.h
```diff
@@ -34,148 +34,148 @@ enum class TypeKind : int8_t {
...
 };
 
 template <TypeKind KIND>
 struct TypeTraits {};
 
 template <>
 struct TypeTraits<TypeKind::kBool> {
-  static constexpr const char* signature = "bool";
-  static constexpr const char* typeString = "boolean";
+  static constexpr const char* kSignature = "bool";
```
EpsilonPrime, line 45: Looks like I should be using these in the converter.  Added to the list.

.clang-format
```diff
@@ -83,5 +83,4 @@ SpacesInParentheses: false
 SpacesInSquareBrackets: false
 Standard:        Cpp11
 TabWidth:        8
-UseTab:          Never
-...
+UseTab:          Never
```
EpsilonPrime, line 86: Could we add a trailing linefeed at the end of this file (also .clang-tidy and third-party/.clang-tidy).

src/substrait/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
-string(TOLOWER ${CMAKE_BUILD_TYPE} BUILD_SUBDIR_NAME)
+#string(TOLOWER ${CMAKE_BUILD_TYPE} BUILD_SUBDIR_NAME)
```
EpsilonPrime, line 3: If this line is not needed could we just remove it?

src/substrait/textplan/Any.h
```diff
@@ -9,7 +9,7 @@
 namespace io::substrait::textplan {
 
 template <class ValueType>
-inline ValueType any_cast(const std::any& value, const char* file, int line) {
+inline ValueType AnyCast(const std::any& value, const char* file, int line) {
```
EpsilonPrime, line 12: Since this function wraps std::any_cast it seemed appropriate to use the same style.

src/substrait/textplan/ParseResult.h
```diff
@@ -18,42 +18,41 @@ class ParseResult {
       SymbolTable symbolTable,
       std::vector<std::string> syntaxErrors,
       std::vector<std::string> semanticErrors) {
-    symbol_table_ = std::make_shared<SymbolTable>(std::move(symbolTable));
```
EpsilonPrime, line 21: Class data members in Google C++ style use underscores.  Have we documented the style Substrait C++ that I missed?

EpsilonPrime (approved):
I'm okay with making all of these changes but we should make sure we're all following the same style guide (and I'm okay with whatever style guide as long as I know what it is).

.clang-format
chaojun-zhang: Yes

src/substrait/textplan/ParseResult.h
chaojun-zhang: Even the google source code doesn't follow the google c++ style strictly , see  [this](https://github.com/googleapis/google-cloud-cpp/blob/main/.clang-tidy)

westonpace:
> > I'm okay with making all of these changes but we should make sure we're all following the same style guide (and I'm okay with whatever style guide as long as I know what it is).
> 
> I couldn't agree more, but I didn't find a style that close to the code in this project, I try my best to make our style uniform and change less code, the original code style is follow the facebook C++ coding style.

I also agree that consistency is more important than any particular detail.  It appears that many of these things are being enforced now by either clang-format or clang-tidy?

If so, then the `.clang-tidy` and `.clang-format` files are a style guide of sorts.  I think the only things we need to write down beyond that are things that clang-tidy does not enforce.

> #  -performance-move-const-arg: This warning requires the developer to
> #      know/care more about the implementation details of types/functions than
> #      should be necessary. For example, `A a; F(std::move(a));` will trigger a
> #      warning IFF `A` is a trivial type (and therefore the move is
> #      meaningless). It would also warn if `F` accepts by `const&`, which is
> #      another detail that the caller need not care about.

This is interesting and new to me.  Although I can see how it would make things simpler.

> #  -bugprone-narrowing-conversions: too many false positives around
> #      `std::size_t`  vs. `*::difference_type`.

Personally, I prefer warning on this and using a lot of explicit casts.  It could help to avoid subtle bugs.  However, if we want to skip these I think it's a pretty common thing to do.

chaojun-zhang:
> If so, then the `.clang-tidy` and `.clang-format` files are a style guide of sorts. I think the only things we need to write down beyond that are things that clang-tidy does not enforce.
 

Currently the `readability-identifier-naming.FunctionCase` and `readability-identifier-naming.VariableCase` both set as  `aNy_CasE ` which is not consistence with google style , so they are not enforce by clang-tidy check, we  need to  make agreement for them

westonpace:
> Following case style are inconsistence with google c++ style :
> 
>     * readability-identifier-naming.FunctionCase(Not enforced)
> 
>     * readability-identifier-naming.VariableCase(Not enforced)
> 
>     * readability-identifier-naming.ClassMemberCase (change to camelBack, default is lower_case)
> 
> 
> We need to make agreement for them

I agree we should pick something.  I will defer to @chaojun-zhang and @EpsilonPrime if either of you have string opinions.  Otherwise, my understanding is that we have been using:


readability-identifier-naming.FunctionCase: camelCase
readability-identifier-naming.VariableCase: camelCase
readability-identifier-naming.ClassMemberCase: camelCase (with trailing underscore?)

EpsilonPrime:
Definitely fine with going with what we've got in the codebase so far.

.cmake-format
```diff
@@ -0,0 +1,7 @@
+# How many spaces to tab for indent
+tab_size: 2
+# Format command names consistently as 'lower' or 'upper' case
+command_case: "lower"
+first_comment_is_literal: False
+# enable comment markup parsing and reflow
+enable_markup: False
```
EpsilonPrime, line 7: Missing line at the end of file.

.github/workflows/build_test.yml
```diff
@@ -39,4 +42,4 @@ jobs:
       - name: Build
         run: ninja -C build
       - name: Test
-        run: ctest --test-dir build --output-on-failure
+        run: ctest --test-dir build --output-on-failure
```
EpsilonPrime, line 45: Missing line at the end of file.

scripts/setup-ubuntu.sh
```diff
@@ -24,6 +17,7 @@ sudo --preserve-env apt install -y \
   wget \
   libprotobuf-dev \
   libprotobuf23 \
-  protobuf-compiler
+  protobuf-compiler \
+  clang-format
 
 pip install cmake-format
```
EpsilonPrime, line 23: Missing linefeed at end of file.

src/substrait/textplan/tests/ParseResultMatchers.h
```diff
@@ -9,20 +9,20 @@
 
 namespace io::substrait::textplan {
 
-[[maybe_unused]] ::testing::Matcher<const ParseResult&> ParsesOk();
```
EpsilonPrime, line 12: The style here matches that of the rest of the Googletest matchers.  If it's easier to do, we can use our consistent case but I'd prefer to keep these in line if possible.  For instance, mixing whenSerialized with HasSubstr.

whenSerialized(HasSubstr("text"))

EpsilonPrime (commented):
LGTM

.github/workflows/build_test.yml
EpsilonPrime: We probably want to put this in another PR but here's a workflow that enforces files end in newlines:

```
 - name: All files end with newline
 - run: find dir1 dir2 -type f | xargs -L1 bash -c 'test "$(tail -c1 "$0")" && echo "No newline at end of $0"'
```

Here's the article:
https://medium.com/@alexey.inkin/how-to-force-newline-at-end-of-files-and-why-you-should-do-it-fdf76d1d090e

src/substrait/textplan/tests/ParseResultMatchers.h
chaojun-zhang: Not easy to do that, we have to add ` // NOLINT`  for all of them to keep in line with google test matchers.  I saw your other's PR all follow google test style.  if we have to do that, then we  don't need to enforce and change the function case style back to `aNy_Case`, what's your opinion?

src/substrait/textplan/converter/tests/BinaryToTextPlanConversionTest.cpp
```diff
@@ -175,14 +176,17 @@ std::vector<TestCase> GetTestCases() {
...
+          "relations: { root: { input: { project: { input { read: { "
+          "local_files {} } } } } } }",
+          hasSymbols({"local", "read", "project", "root"}),
       },
       {
           "two identical three node pipelines",
-          "relations: { root: { input: { project: { input { read: { local_files {} } } } } } }"
-          "relations: { root: { input: { project: { input { read: { local_files {} } } } } } }",
-          HasSymbols(
+          "relations: { root: { input: { project: { input { read: { "
```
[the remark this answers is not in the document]
chaojun-zhang: Fixed

.github/workflows/build_test.yml
chaojun-zhang: I will added to list

westonpace (approved):
Unless there are any major changes let's go ahead and merge this soon.  We can fix minor things in follow-up.  That way we can move forward on some of the other PRs without having to worry about merge conflicts.  Thanks again for putting this together.

EpsilonPrime (approved):
Yep, all the recent changes LGTM.

Issue #41, opened 2023-03-31
Disable warnings about usage of deprecated `args`


westonpace:
The converter is tolerant of older plans.  This means it will visit the `args` variant of scalar function arguments which is deprecated.  However, this generates compiler warnings.  We should suppress those warnings with the appropriate pragmas.

PR #43, opened 2023-04-03
GH-41: Disable warnings about usage of deprecated args

EpsilonPrime:
The converter is tolerant of older plans so it needs to utilize the deprecated plan format to function properly.

### DIFF
--- a/src/substrait/textplan/converter/BasePlanProtoVisitor.cpp
+++ b/src/substrait/textplan/converter/BasePlanProtoVisitor.cpp
@@ -10,6 +10,8 @@
 #include "substrait/proto/algebra.pb.h"
 #include "substrait/proto/plan.pb.h"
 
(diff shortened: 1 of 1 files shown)


src/substrait/textplan/converter/BasePlanProtoVisitor.cpp
```diff
@@ -10,6 +10,11 @@
 #include "substrait/proto/algebra.pb.h"
 #include "substrait/proto/plan.pb.h"
 
+#pragma clang diagnostic push
```
westonpace, line 13:
```suggestion
// The visitor should try and be tolerant of older plans.  This
// requires calling deprecated APIs.
#pragma clang diagnostic push
```

westonpace (changes requested):
One small suggestion

westonpace (approved):

PR #48, opened 2023-04-11
feat: join relation and proper function scoping

EpsilonPrime:
* Implemented printing of the join relation.
* Now collects the schema for each relation in RelationData.
* Updated field references to be scoped as documented in the Substrait main repo.
* Added simple readme to explain how to use the plan converter tool.

### DIFF
--- a/src/substrait/textplan/converter/InitialPlanProtoVisitor.h
+++ b/src/substrait/textplan/converter/InitialPlanProtoVisitor.h
@@ -5,6 +5,7 @@
 #include <any>
 
 #include "substrait/proto/plan.pb.h"
+#include "substrait/textplan/RelationData.h"
 #include "substrait/textplan/SubstraitErrorListener.h"
 #include "substrait/textplan/SymbolTable.h"
 #include "substrait/textplan/converter/BasePlanProtoVisitor.h"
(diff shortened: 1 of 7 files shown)


src/substrait/textplan/converter/InitialPlanProtoVisitor.h
```diff
@@ -56,6 +57,28 @@ class InitialPlanProtoVisitor : public BasePlanProtoVisitor {
...
+      const ::substrait::proto::Rel& right);
+
+  template <typename T>
+  void addFieldsToRelation(
+      const std::shared_ptr<RelationData>& relationData,
+      const T& relations) {
+    for (const auto& rel : relations) {
+      addFieldsToRelation(relationData, rel);
+    }
+  };
```
westonpace, line 80: This is a little generic.  I'm not sure there is much we can do about it.  Ideally we'd be able to restrict it to "things that can be ranged over" but without concepts (e.g. https://en.cppreference.com/w/cpp/ranges/range) I don't think we can easily do that.

westonpace (approved):
One small nit but probably not worth upgrading to C++20 over.

src/substrait/textplan/converter/InitialPlanProtoVisitor.h
EpsilonPrime: Agreed, I did look at ranges when I was implementing this.

PR #51, opened 2023-04-15
refactor:  Separate type signature conversion into its own module

EpsilonPrime:
Moves proto type (not the in-memory type Type) to type signature conversion routine into its own module.  The reverse conversion will join this routine in an upcoming PR.

### DIFF
--- a/src/substrait/textplan/TypeConversion.cpp
+++ b/src/substrait/textplan/TypeConversion.cpp
@@ -6,127 +6,134 @@
 
 namespace io::substrait::textplan {
 
+namespace {
+
+std::string kNullSignature = "?";
+
+} // namespace
+
 // Converts a Type proto directly into its text signature form.
 std::string typeToText(const ::substrait::proto::Type& type) {
   switch (type.kind_case()) {
     case ::substrait::proto::Type::kBool:
       if (type.bool_().nullability() ==
           ::substrait::proto::Type::NULLABILITY_NULLABLE) {
-        return "bool?";
+        return TypeTraits<TypeKind::kBool>::kSignature + kNullSignature;
       }
       return TypeTraits<TypeKind::kBool>::kSignature;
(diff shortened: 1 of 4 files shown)


src/substrait/textplan/TypeConversion.cpp
```diff
@@ -0,0 +1,151 @@
...
+namespace io::substrait::textplan {
+
+namespace {
+
+std::string kNullSignature = "?";
+
+} // namespace
+
+// Converts a Type proto directly into its text signature form.
+std::string typeToText(const ::substrait::proto::Type& type) {
```
chaojun-zhang, line 16: how about refactor this method as  `std::share_ptr<ParameterType> fromSubstraitType(const ::substrait::proto::Type& type)` and move into header `Type.h` , and then we can get the signature via  ParameterType::signature() ?

EpsilonPrime: I considered including the methods there but I was concerned about adding the protobufs as a dependency.  If this isn't an issue I'm up for moving this there.  @chaojun-zhang

src/substrait/textplan/TypeConversion.cpp
```diff
@@ -0,0 +1,151 @@
...
+  switch (type.kind_case()) {
+    case ::substrait::proto::Type::kBool:
+      if (type.bool_().nullability() ==
+          ::substrait::proto::Type::NULLABILITY_NULLABLE) {
+        return TypeTraits<TypeKind::kBool>::kSignature + kNullSignature;
```
chaojun-zhang, line 21: why use  kSignature as text string  for bool type here and kTypeString for other types?

EpsilonPrime:
I have a PR in progress that will add fromProto to the Substrait type module.  Then a followup PR to that will update PlanPrinterVisitor.cpp (the current location of this code) to use that functionality.

PR #53, opened 2023-04-19
fix: Correct nullable handling behavior for complex types

EpsilonPrime:
Previously a question mark anywhere within a type string would make a type nullable.  This PR corrects that calculation.

### DIFF
--- a/src/substrait/type/tests/TypeTest.cpp
+++ b/src/substrait/type/tests/TypeTest.cpp
@@ -206,6 +206,14 @@ TEST_F(TypeTest, decodeTest) {
         ASSERT_TRUE(typePtr->elementType()->isPlaceholder());
       });
 
+  testDecode<ParameterizedList>(
+      "list<string?>",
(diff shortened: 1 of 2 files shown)


src/substrait/type/tests/TypeTest.cpp
```diff
@@ -192,6 +192,14 @@ TEST_F(TypeTest, decodeTest) {
         ASSERT_EQ(typePtr->signature(), "list<str>");
       });
 
+  testDecode<ParameterizedList>(
+      "LIST<any>?",
```
chaojun-zhang, line 196: Seems question mark can only appear prior to  the position of parameter type , see https://substrait.io/types/type_parsing/#type-syntax-parsing

EpsilonPrime: I've removed the decoding enhancement for now.  The intent was to make the decoding more lenient but that should be done as part of a separate PR from this bug fix.

chaojun-zhang (approved):

westonpace (approved):

PR #56, opened 2023-04-28
feat:  Add first mismatch to EqSquashingWhitespace googletest matcher

EpsilonPrime:
Now emits the first mismatch between the two strings to aid in determining what went wrong.

This PR also updates the parameter passing style used.


westonpace (approved):
Thanks for the extra diagnostics.

PR #59, opened 2023-05-09
feat(refactor): Use symbols instead of names for field references

EpsilonPrime:


src/substrait/textplan/StructuredSymbolData.h
```diff
@@ -25,8 +27,7 @@ struct RelationData {
   std::vector<const ::substrait::proto::Rel*> newPipelines;
 
   // Column name for each field known to this relation (in field order).
-  // TODO -- Use symbols instead of column names here.
-  std::vector<std::string> fieldReferences;
+  std::vector<const SymbolInfo*> fieldReferences;
```
westonpace, line 30: Are these input fields or output fields?  If a relation has multiple inputs then how are these distributed?

westonpace (approved):
Maybe let's just add a clarifying comment and then we're good to go.

src/substrait/textplan/StructuredSymbolData.h
EpsilonPrime: Added an indication of their current use (both input and output for now).

PR #62, opened 2023-05-18
fix: Prevent crash when given a map type with no key or value specified

EpsilonPrime:


src/substrait/type/Type.cpp
```diff
@@ -527,7 +527,9 @@ std::string ParameterizedMap::signature() const {
   sign << "<";
   sign << keyType()->signature();
```
westonpace, line 528: Do we need to worry about `keyType()` being `nullptr`?

westonpace (approved):
Just a question, but otherwise this seems fine.  We should probably avoid crashes in any case.

src/substrait/type/Type.cpp
EpsilonPrime: Check added.

PR #65, opened 2023-06-01
fix: detect when maps are missing keys and/or values

EpsilonPrime:
Now that the underlying type library doesn't segfault when given a map missing a key or value, properly detect such cases and return errors.


westonpace (approved):
I agree these are better error messages.  Thanks!

PR #66, opened 2023-06-02
feat: fields, schemas, and sources binary output for the textplan parser

EpsilonPrime:
* Fields are now emitted in expressions completing expression handling.
    * Schemas and sources are also now implemented.
    * Removes projection as a relation detail in favor of expression for project relations.

### DIFF
--- a/src/substrait/textplan/parser/SubstraitPlanRelationVisitor.cpp
+++ b/src/substrait/textplan/parser/SubstraitPlanRelationVisitor.cpp
@@ -3,6 +3,7 @@
 #include "substrait/textplan/parser/SubstraitPlanRelationVisitor.h"
 
 #include <chrono>
+#include <limits>
 #include <memory>
 #include <sstream>
 #include <string>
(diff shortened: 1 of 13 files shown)


src/substrait/textplan/parser/SubstraitPlanVisitor.cpp
```diff
@@ -335,10 +351,10 @@ std::any SubstraitPlanVisitor::visitLocal_files_detail(
   for (const auto& f : ctx->file()) {
     symbolTable_->defineSymbol(
         f->getText(),
-        Location(ctx),
+        PARSER_LOCATION(ctx->parent->parent),
```
westonpace, line 354: `ctx->parent->parent` is a little confusing.  Can you add a comment here?

src/substrait/textplan/parser/SubstraitPlanRelationVisitor.cpp
```diff
@@ -435,7 +521,25 @@ std::any SubstraitPlanRelationVisitor::visitExpressionConstant(
...
+      [&](auto ref) { return (ref->name == symbolName); });
+
   ::substrait::proto::Expression expr;
+  if (currentFieldNumber != relationData->fieldReferences.end()) {
+    expr.mutable_selection()
+        ->mutable_direct_reference()
+        ->mutable_struct_field()
+        ->set_field(static_cast<int32_t>(
+            (currentFieldNumber - relationData->fieldReferences.begin()) &
+            0xffff));
```
westonpace, line 540: Why `& 0xFFFF`?  Should we assert that this value is less than 2 bytes?

westonpace (approved):
Just two minor things that I noticed.  Otherwise, I like the way this is starting to shape up

src/substrait/textplan/parser/SubstraitPlanRelationVisitor.cpp
EpsilonPrime: Changed to use std::numeric_limits<int32_t>::max

westonpace (approved):

PR #72, opened 2023-06-16
feat:  add ifthen expression support

EpsilonPrime:
In this implementation IFTHEN is being treated as a built-in function.

### DIFF
--- a/src/substrait/textplan/converter/PlanPrinterVisitor.cpp
+++ b/src/substrait/textplan/converter/PlanPrinterVisitor.cpp
@@ -442,10 +442,33 @@ std::any PlanPrinterVisitor::visitWindowFunction(
 
 std::any PlanPrinterVisitor::visitIfThen(
     const ::substrait::proto::Expression::IfThen& ifthen) {
-  errorListener_->addError(
-      "If then expressions are not yet supported: " +
-      ifthen.ShortDebugString());
-  return std::string("IFTHEN_NOT_YET_IMPLEMENTED");
+  std::stringstream text;
+  text << "IFTHEN(";
+  bool hasPreviousText = false;
+  for (const auto& clause : ifthen.ifs()) {
+    if (!clause.has_if_() || !clause.has_then()) {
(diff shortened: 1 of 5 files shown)


src/substrait/textplan/converter/PlanPrinterVisitor.cpp
```diff
@@ -442,10 +442,28 @@ std::any PlanPrinterVisitor::visitWindowFunction(
...
+  std::stringstream text;
+  text << "IFTHEN(";
+  bool hasPreviousText = false;
+  for (const auto& clause : ifthen.ifs()) {
+    if (hasPreviousText) {
+      text << ", ";
+    }
+    text << ANY_CAST(std::string, visitExpression(clause.if_()));
+    text << ", ";
+    text << ANY_CAST(std::string, visitExpression(clause.then()));
```
westonpace, line 454: What happens if `then` is not set?  Do we crash?  It's ok if it's invalid.

src/substrait/textplan/converter/PlanPrinterVisitor.cpp
```diff
@@ -442,10 +442,28 @@ std::any PlanPrinterVisitor::visitWindowFunction(
...
+    hasPreviousText = true;
+  }
+  if (ifthen.has_else_()) {
+    if (hasPreviousText) {
+      text << ", ";
+    }
+    text << ANY_CAST(std::string, visitExpression(ifthen.else_()));
+    hasPreviousText = true;
```
westonpace, line 462: Super minor nit, could remove.
```suggestion
```

westonpace (changes requested):
As discussed offline, maybe add one small test case outside the round trip tests.

src/substrait/textplan/converter/PlanPrinterVisitor.cpp
EpsilonPrime: Added check and test for the same.

westonpace (approved):

PR #74, opened 2023-06-21
feat: add binary plan normalizer

EpsilonPrime:
The normalizer orders the extension uri spaces and function extensions alphabetically (and renumbers accordingly) so that it is easier to compare binary plans.   A slight change to the existing sort behavior is made to be consistent with this ordering.

### DIFF
--- a/src/substrait/textplan/SymbolTablePrinter.cpp
+++ b/src/substrait/textplan/SymbolTablePrinter.cpp
@@ -254,7 +254,15 @@ std::string outputFunctionsSection(const SymbolTable& symbolTable) {
   std::stringstream text;
 
   std::map<uint32_t, std::string> spaceNames;
-  std::set<uint32_t> usedSpaces;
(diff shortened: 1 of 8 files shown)


src/substrait/textplan/SymbolTablePrinter.cpp
```diff
@@ -254,7 +254,15 @@ std::string outputFunctionsSection(const SymbolTable& symbolTable) {
   std::stringstream text;
 
   std::map<uint32_t, std::string> spaceNames;
-  std::set<uint32_t> usedSpaces;
+  auto cmp = [&](uint32_t a, uint32_t b) {
```
westonpace, line 257: Maybe add a small comment just explaining why we are doing this

westonpace (approved):
Looks good.  Probably don't need to dive too far into test internals too.  It makes sense you'd want a canonical ordering.

PR #76, opened 2023-06-23
feat:  add root names to the textplan

EpsilonPrime:
Root names are added to both the parser and converter in this PR.  The root relation is not being treated as a relation internally and instead is treated merely as annotation containing a list of names.  This is primarily because the relation-related codepaths make assumptions that wouldn't apply to the root (such as having a valid Relation proto as its data type).


westonpace (changes requested):
Can we add one example plan with a root in it for text->binary

westonpace (approved):

PR #77, opened 2023-06-24
refactor:  upgrade multi-key sort

EpsilonPrime:
As suggested by @bkietz this PR upgrades the sort routine to utilize tuple's built-in compare behavior.  This reduces the complexity of evaluating each set of four possibilities for each key manually.

### DIFF
--- a/src/substrait/textplan/converter/ReferenceNormalizer.cpp
+++ b/src/substrait/textplan/converter/ReferenceNormalizer.cpp
@@ -14,27 +14,19 @@ namespace {
 bool compareExtensionFunctions(
     const ::substrait::proto::extensions::SimpleExtensionDeclaration& a,
     const ::substrait::proto::extensions::SimpleExtensionDeclaration& b) {
-  // First sort so that extension functions proceed any other kind of extension.
-  if (a.has_extension_function() && !b.has_extension_function()) {
-    return true;
-  } else if (!a.has_extension_function() && b.has_extension_function()) {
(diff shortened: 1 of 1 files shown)


src/substrait/textplan/converter/ReferenceNormalizer.cpp
```diff
@@ -14,27 +14,19 @@ namespace {
...
-  return a.extension_function().name() < b.extension_function().name();
+  auto ord = [](const auto& decl) {
+    return make_tuple(
+        // First sort so that extension functions precede any other kind of
+        // extension.
+        not decl.has_extension_function(),
+        // Next sort by space.
+        decl.extension_function().extension_uri_reference(),
+        // Finally sort by name within a space.
+        decl.extension_function().name());
```
bkietz, line 25: Using `make_tuple` here will create copies of the strings on each comparison. We can avoid this by wrapping the strings in `string_view`
```suggestion
    return make_tuple(
        // First sort so that extension functions precede any other kind of
        // extension.
        not decl.has_extension_function(),
        // Next sort by space.
        std::string_view{decl.extension_function().extension_uri_reference()},
        // Finally sort by name within a space.
        std::string_view{decl.extension_function().name()});
```

bkietz (changes requested):

westonpace (approved):
Seems quite a bit simpler.  I'm not sure how I feel about `not` vs. `!` but I don't want to get hung up on these minor details.

PR #79, opened 2023-07-08
fix:  support both old and new versions of protobuf

EpsilonPrime:
The version numbering scheme for the protobuf compiler was updated which causes a mismatch warning to appear for versions of cmake less than 3.27.  Later versions of protobuf are bundled with their own copy of abseil.  Since protobuf always required abseil anyway, this fix also updates the code to rely on that (when available) instead of our own submodule.  This should lead to greater consistency (especially with new protobuf code leaning on an experimental logging feature that won't be available in older versions of abseil).  If the newer CONFIG-based method of findpackage does not work, we now fall back to the older method of using findpackage.  The version of cmake that we require is still 3.20 as 3.27 isn't available in enough locations to advance to that minimum.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,12 +16,21 @@ option(
   "Enable substrait-cpp tests. This will enable all other build options automatically."
   ON)
 
-find_package(Protobuf REQUIRED)
-include_directories(${PROTOBUF_INCLUDE_DIRS})
-
-add_subdirectory(third_party)
+# Local files come first.
 include_directories(include)
 include_directories(src)
+
+add_subdirectory(third_party)
+
+#find_package(absl REQUIRED)
+#include_directories(${ABSL_INCLUDE_DIRS})
+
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
```diff
@@ -16,13 +16,18 @@ option(
...
-add_subdirectory(third_party)
+# Local files come first.
 include_directories(include)
 include_directories(src)
 
+add_subdirectory(third_party)
+
+find_package(Protobuf QUIET CONFIG)
+if(NOT ${Protobuf_FOUND})
+  find_package(Protobuf REQUIRED)
```
westonpace, line 27: Can we add a comment explaining why we need to do this?

westonpace (changes requested):
One minor comment and I think we are good here.

CMakeLists.txt
EpsilonPrime: Done.

westonpace (approved):

PR #81, opened 2023-08-11
feat: enable roundtrip testing of textplan conversion

EpsilonPrime:


src/substrait/textplan/tests/CMakeLists.txt
```diff
@@ -19,50 +19,38 @@ add_test_case(
...
-
-  message(
-    STATUS
-      "Round trip testing is turned off.  Add SUBSTRAIT_CPP_ROUNDTRIP_TESTING=on to enable."
-  )
+add_test_case(
+  substrait_textplan_round_trip_test
+  SOURCES
+  RoundtripTest.cpp
+  EXTRA_LINK_LIBS
```
mbrobbel, line 26: Just one question about `EXTRA_LINK_LIBS`: what is it and why is it in the sources list?

EpsilonPrime: Good question -- it's actually another argument (section keyword) from the cmake function we use to make our tests:

[BuildUtils.cmake](https://github.com/substrait-io/substrait-cpp/blob/main/cmake_modules/BuildUtils.cmake#L32.)

mbrobbel (approved):

PR #98, opened 2024-03-04
feat: add plantransformer tool to convert any plan to any format

EpsilonPrime:


src/substrait/common/Tool.cpp
mbrobbel, line 1: Nit: maybe rename this file to `PlanTransformerTool.cpp`?

mbrobbel (approved):

Issue #99, opened 2024-03-13
Break cyclic dependencies of CMake targets


ingomueller-net:
There seem to be several cycles in the dependency graph of CMake targets. This isn't a problem for compiling static libraries, but if I build with `cmake -DBUILD_SHARED_LIBS=ON ...`, I get the following error (uninteresting parts omitted):

```
CMake Error: The inter-target dependency graph contains the following strongly connected component (cycle):
  "substrait_io" of type STATIC_LIBRARY
    depends on "substrait_textplan_converter" (weak)
    depends on "symbol_table" (weak)
    depends on "substrait_textplan_loader" (weak)
    depends on "substrait_textplan_converter" (strong)
    depends on "substrait_textplan_loader" (strong)
  "symbol_table" of type SHARED_LIBRARY
    depends on "substrait_textplan_converter" (weak)
    depends on "substrait_io" (weak)
    depends on "substrait_textplan_loader" (weak)
  "substrait_textplan_converter" of type SHARED_LIBRARY
    depends on "substrait_io" (weak)
    depends on "substrait_textplan_loader" (weak)
    depends on "symbol_table" (weak)
  "substrait_textplan_loader" of type SHARED_LIBRARY
    depends on "symbol_table" (weak)
    depends on "substrait_textplan_converter" (weak)
    depends on "substrait_io" (weak)
At least one of these targets is not a STATIC_LIBRARY.  Cyclic dependencies are allowed only among static libraries.
```

I understand that there are reasons not to build shared libraries, but I argue that there are situations where it is useful. (I am working on a project with many top-level binaries that share 99% of the code and compilation and disk usage is *a lot* better with shared libs, so I hugely prefer that for developing and CI.)

I am currently trying to consume `substrait-cpp` as a dependency using `add_subdirectory`. Since I build my top-level project with `BUILD_SHARED_LIBS`, `substrait-cpp` is built with the same configuration. For any scenario where that flag is desired or required for the top-level project, support that flag would be great.

Finally, I believe that cyclic dependencies are often an indication for sloppy design or code organization and therefor a good idea to clean up. However, I haven't looked into the cycles yet, so I don't know yet whether there are good reasons to have them or things have just grown this way...

ingomueller-net:
OK, one cycle can be broken very easily:

```diff
diff --git a/src/substrait/textplan/converter/CMakeLists.txt b/src/substrait/textplan/converter/CMakeLists.txt
index 5ee2753..f328f81 100644
[...]

PR #100, opened 2024-03-14
fix: correct cmake library interdependencies

EpsilonPrime:
This eliminates the cycles described in #99.

PlanPrinterVisitor is now part of symbol_table.
BasePlanProtoVisitor (which PlanPrinterVisitor depends on) is now its own library.


ingomueller-net:
Awesome -- thank you for the quick fix! Probably a much better fix than what I had in mind :)

For completeness, here are the dependency graphs before and after this PR. Before the change, we can see several inter-dependent targets; after the change, it's a nice DAG.

**Before:**

before

**After:**

after

I generated them running `cmake --graphviz=foo.dot .. -DBUILD_SHARED_LIBS=OFF && dot -Tpng -o foo.png foo.dot` in the build folder, where I put a file called `CMakeGraphVizOptions.cmake` with the following content

```cmake
set(GRAPHVIZ_IGNORE_TARGETS "absl_*;gmock*;gtest*;protobuf*;yaml-cpp*;fmt*;-lrt;Threads*;/usr/*;.*_test"
```

westonpace (approved):
Approve if CI passes.  I won't pretend to have fully internalized these changes but I don't see anything odd or out of place.

EpsilonPrime:
Thanks for finding this issue as well as providing the visualization of the dependencies.  I'll probably use this trick to prevent such issues in the future.

ingomueller-net:
> Thanks for finding this issue as well as providing the visualization of the dependencies. I'll probably use this trick to prevent such issues in the future.

Sure thing! Another possibility is to (also) build with `BUILD_SHARED_LIBS` in CI...

PR #101, opened 2024-03-14
fix: Set TEXTPLAN_SOURCE_DIR based on current dir in planloader test.

ingomueller-net:
That variable was previously computed based on `CMAKE_SOURCE_DIR`, which is the wrong value if `substrait-cpp` is used as a dependency from a different top-level CMake project. The PR applies the pattern based on `CMAKE_CURRENT_SOURCE_DIR` used in other tests.

### DIFF
--- a/export/planloader/tests/CMakeLists.txt
+++ b/export/planloader/tests/CMakeLists.txt
@@ -15,10 +15,7 @@ add_test_case(
   gtest
   gtest_main)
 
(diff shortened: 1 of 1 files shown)


westonpace (approved):

ingomueller-net:
Thanks for the prompt approvals! Note that I don't have permissions to merge. Feel free to merge at any time...

EpsilonPrime:
Yep, was just waiting for my CI fixes to catch up so the tests would pass.  Thanks again.  There hasn't been a lot of use of this as a sub project yet.

PR #103, opened 2024-03-15
feat: Make main target installable.

ingomueller-net:
This allows to run `cmake --install .` to install the project to a
system such that it can be consumed with `find_package()`. This required
the following changes:

* Add `install(TARGET ...)` commands for all library and executable targets.
* Add all public header files to the `HEADER` `FILE_SET` of their respective target.
* Export all these targets to an export set and `install(EXPORT ...)` that export set.
* Provide install-time paths for several `target_include_directories()`,
  some of which are empty because the headers are not installed.

Since some of the library targets depend on external projects (`abseil-cpp`
and `yaml-cpp`), the install logic of these projects also needs to be enabled.

This change is not only necessary to actually install the project, but
also to consume it via `add_subdirectory` from any project that,
itself, is installable. (All dependencies of installable targets must be
installable.)

### DIFF
--- a/src/substrait/proto/CMakeLists.txt
+++ b/src/substrait/proto/CMakeLists.txt
@@ -79,7 +79,9 @@ endforeach()
 # Add the generated protobuf C++ files to our exported library.
 add_library(substrait_proto ${PROTO_SRCS} ${PROTO_HDRS} ProtoUtils.cpp
                             ProtoUtils.h)
-set_target_properties(substrait_proto PROPERTIES PUBLIC_HEADER "${PROTO_HDRS}")
+target_sources(
+  substrait_proto PUBLIC FILE_SET HEADERS BASE_DIRS
+                         ${PROTO_OUTPUT_TOPLEVEL_DIR} FILES ${PROTO_HDRS})
 
 # Include the protobuf library as a dependency to use this class.
 target_link_libraries(substrait_proto protobuf::libprotobuf)
@@ -95,5 +97,5 @@ target_include_directories(
 install(
   TARGETS substrait_proto
   EXPORT SubstraitTargets
-  LIBRARY
-  PUBLIC_HEADER DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/substrait/proto")
+  LIBRARY FILE_SET HEADERS
+          DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/substrait/proto")
--- a/src/substrait/proto/CMakeLists.txt
+++ b/src/substrait/proto/CMakeLists.txt
@@ -86,5 +86,7 @@ target_link_libraries(substrait_proto protobuf::libprotobuf)
 # Make sure we can see our own generated include files.
 target_include_directories(
   substrait_proto
-  PUBLIC "${PROTO_OUTPUT_TOPLEVEL_DIR}/src"
-  PUBLIC "${protobuf_SOURCE_DIR}/src")
+  PUBLIC $<BUILD_INTERFACE:${PROTO_OUTPUT_TOPLEVEL_DIR}/src>
+         $<INSTALL_INTERFACE:include>
+  PUBLIC $<BUILD_INTERFACE:${protobuf_SOURCE_DIR}/src>
+         $<INSTALL_INTERFACE:include>)
--- a/src/substrait/proto/CMakeLists.txt
+++ b/src/substrait/proto/CMakeLists.txt
@@ -77,11 +77,17 @@ foreach(PROTO_FILE IN LISTS PROTOBUF_FILELIST)
 endforeach()
 
 # Add the generated protobuf C++ files to our exported library.
-add_library(substrait_proto ${PROTO_SRCS} ${PROTO_HDRS} ProtoUtils.cpp
-                            ProtoUtils.h)
+add_library(substrait_proto ${PROTO_SRCS} ${PROTO_HDRS} ProtoUtils.cpp)
 target_sources(
-  substrait_proto PUBLIC FILE_SET HEADERS BASE_DIRS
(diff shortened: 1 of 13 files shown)


ingomueller-net:
Note that this PR assumes that we only want to install the functions from the `planloader` targets. In particular, this means that all header files except `Io.hpp` are not installed and thus not available to external projects. I didn't initially understand that (assumed) intention and exported all files, see [this branch](https://github.com/ingomueller-net/substrait-cpp/tree/install-targets).

While implementing this, I realized that most targets have a `substrait_` prefix but not all. In particular, the following of those that are exported don't:

* `planloader`
* `symbol_table`
* `error_listener`
* `textplan_grammar`

Maybe those (and the non-exported ones) should be renamed? I think it is most important for the exported ones since for example `libsymbol_table.a` might collide with other libraries.

ingomueller-net:
Let me also share how I tested this:

```bash
cmake --build . \
    && rm -rf /tmp/substrait-cpp-install/ \
    && cmake --install . --prefix /tmp/substrait-cpp-install \
    && find /tmp/substrait-cpp-install/ \
    | sort | sed "s~/tmp/substrait-cpp-install/~~" \
    | grep -v libabsl | grep -iE "(substrait|lib/lib)"
```

This spits out a (filtered) list of files that are installed and eyeballing at that list suggest that all necessary files are there. It would be better, though, to actually build a project that consumes these files and ideally build that project in CI.

ingomueller-net:
I just realized that this doesn't take care of the protobuf headers, which are included from Io.hpp. Will look at that later.

EpsilonPrime:
The thought behind the export directory was for sharing simplified interfaces for Java and Python wrapping of the libraries.  substrait/type and substrait/common are the main public interfaces with substrait/textplan as usable with less interference.  To make the distinction of public more clear, anything with its include files under the toplevel include directory is public (instead of local to the module).

I agree the target names should all be updated but we can deal with that in a separate PR.

I'm not sure the default option should be for the YAML library to be exposed (it may be only needed by tests).  Perhaps it ought to be fetched if needed like some of the other required libraries instead of being submodule'd (which is where I've been headed towards lately).  I'll have to think a bit more about that.

Looks good so far with the exception of the YAML failure.

src/substrait/proto/CMakeLists.txt
```diff
@@ -79,12 +79,23 @@ endforeach()
 # Add the generated protobuf C++ files to our exported library.
 add_library(substrait_proto ${PROTO_SRCS} ${PROTO_HDRS} ProtoUtils.cpp
                             ProtoUtils.h)
+target_sources(
+  substrait_proto PUBLIC FILE_SET HEADERS BASE_DIRS
+                         ${PROTO_OUTPUT_TOPLEVEL_DIR} FILES ${PROTO_HDRS})
```
ingomueller-net, line 84: This is a mechanism introduced in CMake v2.23. I found this [SO post](https://stackoverflow.com/questions/54271925/how-to-configure-cmakelists-txt-to-install-public-headers-of-a-shared-library) quite useful. Not sure if we want this. If yes, the CMake version requirements in the top-level CMake file should probably be changed. Otherwise, we could have one `install(FILES ...)` command per file in the loop above.

ingomueller-net:
> The thought behind the export directory was for sharing simplified interfaces for Java and Python wrapping of the libraries. substrait/type and substrait/common are the main public interfaces with substrait/textplan as usable with less interference. To make the distinction of public more clear, anything with its include files under the toplevel include directory is public (instead of local to the module).

OK, that also makes sense. I haven't worked with any Java interfacing, so this didn't come to me.

I should amend the PR and export all header files then.

> I agree the target names should all be updated but we can deal with that in a separate PR.

OK.

> I'm not sure the default option should be for the YAML library to be exposed (it may be only needed by tests). Perhaps it ought to be fetched if needed like some of the other required libraries instead of being submodule'd (which is where I've been headed towards lately). I'll have to think a bit more about that.
> 
> Looks good so far with the exception of the YAML failure.

The `substrait_function` library also depends on the YAML library. How would that work if the YAML library isn't available? That's why I configured the library to be installed; otherwise, dependent targets can't be installed...

(Also, what do you mean exactly with "exposed"? Included/FetchContent'd/... in CMake at all? Or installed/exported?)

EpsilonPrime:
By exposed I ideally wanted to limit the number of libraries/targets that were created and installed.  For instance, the textplan_grammar library is something that's internally separated and I would expect that to be statically included to hide its interface from the rest of the world.  Not that it matters too much as an extra library lying around wouldn't hurt too much and by not providing the header files it's still effectively internal.

We're up to 3.24 for the minimum CMake version at this point so we can use that new mechanism.

ingomueller-net:
I think we won't have a choice: any target that installable targets depend on must be installable as well. (I guess that, in the case of a shared lib depending on a static lib, that wouldn't strictly be required. But it'd be already difficult if someone depends on static lib A that is included in shared lib B. Plus all of that might be different on other OSes.)

OK, good! I think I'll update the other libs to also use the file set mechanism (plus adding the remaining header files).

EpsilonPrime:
Sounds good to me.  Thanks!

ingomueller-net:
OK, I think I am done. Feel free to merge as soon as you approve.

To summarize and extend the possible TODOs from this PR:

- [ ] Rename all libraries (and executables?) to have a `substrait_` prefix.
- [ ] Add an example/testing project that uses the main project after installation and/or using `add_subdirectory` and run that in CI.
- [ ] Consider factoring out the logic for one library into a helper function (such as [`absl_cc_library`](https://github.com/abseil/abseil-cpp/blob/56d3f22/CMake/AbseilHelpers.cmake#L84)) to reduce verbosity and improve maintainability.

EpsilonPrime (approved):

ingomueller-net:
Awesome! Thanks for the assistance and the prompt approvals :)

PR #104, opened 2024-07-18
fix: Use CXX compiler of main project for ANTLR external project.

ingomueller-net:
This PR adds the `-DCMAKE_CXX_COMPILER` flag to the CMake flags of the external project for ANTLR. Without that, the external project uses the system default, which may not be what the user wants (and even be broken).


EpsilonPrime (approved):
Thanks for the update, this was snagged from the Antlr distribution, likely it should be updated there as well.

PR #105, opened 2024-07-26
see if we can update cmake

EpsilonPrime:


jacques-n (approved):
+1

Anytime there is more code deleted than added is a good thing from my pov!

PR #106, opened 2024-07-26
try to add arm to github action

jacques-n:


EpsilonPrime (approved):

jacques-n:
Wow, it took 90m to build on macos

???

EpsilonPrime:
It looked like it built and then stalled waiting for the cleanup to happpen
(between steps).

EpsilonPrime:
It is odd that the required "build" option is still pending and that build was renamed to linux-build.  I don't see an option to change the required builds in the repo though.

jacques-n:
Running again to see if performance continues to be bad.

> It is odd that the required "build" option is still pending and that build was renamed to linux-build. I don't see an option to change the required builds in the repo though.

That's something configured inside the Github UI afaik. I changed the names to be clearer around steps but would need to change at the repo level as well if we decide to merge this.

Will wait to see if perf was a random problem or constant. If constant I'm inclined to shelve this for now.

PR #108, opened 2024-09-03
Use CMake built-in dir create function instead of hard-coded `mkdir`

mortbopet:
... Because all systems may not play nice with `mkdir -p`.


EpsilonPrime (approved):
Thanks, definitely more portable.

PR #109, opened 2024-09-03
Don't concatenate paths in `--proto_path`

mortbopet:
`protoc`'s `--proto_path` argument is expected to be provided multiple times if multiple directories are to be considered. Concatenating with `:` is invalid on some systems (Windows):

```sh
> protoc --help
  -IPATH, --proto_path=PATH   Specify the directory in which to search for
                              imports.  May be specified multiple times;
                              directories will be searched in order.  If not
                              given, the current working directory is used.
                              If not found in any of the these directories,
                              the --descriptor_set_in descriptors will be
                              checked for required proto file.
```


EpsilonPrime (approved):
Thanks!

mortbopet:
@EpsilonPrime thank you for the review; looks like CI is back in the green now.

PR #113, opened 2024-09-05
Don't define member functions inside a nested namespace

mortbopet:
This trips up MSVC, in that member functions aren't properly included in the object files of the various translation units.

e.g. `DumpBin` of `symbol_table.lib`:
image

Generally, this can be fixed by having a `using` declaration in a .cpp file to enable the namespace. From my experience, a general good practice is to never nest definitions in namespaces, unless the namespace is only visible from within a translation unit (e.g. a `detail` namespace inside a .cpp file).

This doesn't fix all of the places in the code where this pattern is used - only the (narrow) compile path that i've been excercising. Someone should probably go through the rest of the codebase and apply this - will be needed if/when a windows CI pipeline is added.

### DIFF
--- a/src/substrait/textplan/SymbolTable.cpp
+++ b/src/substrait/textplan/SymbolTable.cpp
@@ -12,7 +12,7 @@
 #include "substrait/textplan/Location.h"
 #include "substrait/textplan/StructuredSymbolData.h"
 
-namespace io::substrait::textplan {
+using namespace io::substrait::textplan;
 
 const std::string& symbolTypeName(SymbolType type) {
   static std::vector<std::string> names = {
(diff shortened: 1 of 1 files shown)


src/substrait/textplan/SymbolTable.cpp
```diff
@@ -12,7 +12,7 @@
 #include "substrait/textplan/Location.h"
 #include "substrait/textplan/StructuredSymbolData.h"
 
-namespace io::substrait::textplan {
```
EpsilonPrime, line 15: Would changing this to three separate namespace definitions did the problem?

mortbopet: It's hard to find documentation about this (it's sort of a bug-in-limbo...). I'm not at my workstation at the moment so i can't give an answer now, but my hunch would be no.

PR #114, opened 2024-09-06
Don't predeclare a struct as a class

mortbopet:
Holy moly this took a long time to debug.

MSVC gives absolutely zero warnings about if a struct/class is predeclared as one or the other, during compilation.

However, linking also needs to happen. And as it turns out, [whether a thing is a class or struct is part of the MSVC name mangling spec](https://en.wikiversity.org/wiki/Visual_C%2B%2B_name_mangling). Hence, we'd end up with e.g. this function (which references `SymbolInfo`) as being defined in it's object file as the following symbol:
> `lookupSymbolsByLocation@SymbolTable@textplan@substrait@io@@QEBA?AV?$vector@PEBUSymbolInfo@textplan@substrait@io@@V?$allocator@PEBUSymbolInfo@textplan@substrait@io@@@std@@@std@@AEBVLocation@234@@Z`

However, things which reference `SymbolInfo`, but which may have pulled in the incorrect predeclaration prior, will reference it as an incorrect symbol:
> `lookupSymbolsByLocation@SymbolTable@textplan@substrait@io@@QEBA?AV?$vector@PEBVSymbolInfo@textplan@substrait@io@@V?$allocator@PEBVSymbolInfo@textplan@substrait@io@@@std@@@std@@AEBVLocation@234@@Z ...`

e.g. notice the difference between `PEBUSymbolInfo` and `PEBVSymbolInfo`. Enqueue strange linking issues galore.

### DIFF
--- a/src/substrait/textplan/StructuredSymbolData.h
+++ b/src/substrait/textplan/StructuredSymbolData.h
@@ -9,7 +9,7 @@
 
 namespace io::substrait::textplan {
 
-class SymbolInfo;
+struct SymbolInfo;
 
 // Used by the PlanRelation and Relation concepts to track connectivity.
 struct RelationData {


mortbopet:
@EpsilonPrime unsure why CI fails here; this shouldn't change anything wrt. the underlying tests.

EpsilonPrime:
Yeah, really weird as I wouldn't expect changes to file locations to affect the correctness of one test.  I will check into this when I get in.

EpsilonPrime:
The issue is that the timezones "US/Central" and "UTC" are no longer being found in the timezone database.  I am able to reproduce this issue on iOS as well.  Still looking for what changed between today and yesterday.

EpsilonPrime:
Looks like the timezone database was updated yesterday:

https://lists.iana.org/hyperkitty/list/tz-announce@iana.org/thread/IZ7AO6WRE3W3TWBL5IR6PMQUL433BQIE/

EpsilonPrime (approved):
This one's totally my fault.  I'm surprised it wasn't caught by all of the lint and build systems we have.  Thanks for spending the time to find this!

EpsilonPrime:
(The timezone issue appears to be fixed by #120 .  Will merge this after that PR lands.)

PR #117, opened 2024-09-06
Fix namespace pollution on Windows

mortbopet:
As described in https://github.com/substrait-io/substrait-cpp/issues/116, the `io.h` file in this project conflicts with `io.h` of windows.

This fixes that (in the somewhat narrow compilation path that i'm exercising) by removing `include/substrait/common` from the include path, and instead relying on the (already existing) full include-path namespacing instead of relative includes of `io.h`.

### DIFF
--- a/src/substrait/common/CMakeLists.txt
+++ b/src/substrait/common/CMakeLists.txt
@@ -19,10 +19,8 @@ add_dependencies(
   absl::status
   absl::statusor)
 target_include_directories(
-  substrait_io
-  INTERFACE
-    $<BUILD_INTERFACE:${SUBSTRAIT_CPP_INCLUDE_DIR}>
-    $<INSTALL_INTERFACE:include>)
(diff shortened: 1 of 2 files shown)


src/substrait/common/CMakeLists.txt
```diff
@@ -21,8 +21,8 @@ add_dependencies(
 target_include_directories(
   substrait_io
   INTERFACE
-    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../../../include/substrait/common>
-    $<INSTALL_INTERFACE:include/substrait/common>)
+    $<BUILD_INTERFACE:${SUBSTRAIT_CPP_INCLUDE_DIR}>
```
EpsilonPrime, line 24: The linter is complaining here since the lines are so short.  (It wants to use two lines instead of three.)

EpsilonPrime (approved):
Thanks for fixing this.  I was wondering why the common stuff wasn't as accessible as I thought it should have been!

PR #118, opened 2024-09-06
Don't throw away `visit` result in BasePlanProtoVisitor

mortbopet:
`visitPlan` is marked `protected`, meaning that non-subclasses of `BasePlanProtoVisitor` (which is probably what'll initiate the `visit` most of the time) cannot have access to the `visit` result.

Is there a specific reason that the traversal result has been hidden? If not, i suggest this change.

If so, subclasses are obviously free to implement a `visitWithResult` function to do the same thing.

### DIFF
--- a/src/substrait/textplan/converter/BasePlanProtoVisitor.h
+++ b/src/substrait/textplan/converter/BasePlanProtoVisitor.h
@@ -17,8 +17,8 @@ class BasePlanProtoVisitor {
   BasePlanProtoVisitor() = default;
 
   // visit() begins the traversal of the entire plan.
-  virtual void visit(const ::substrait::proto::Plan& plan) {
-    visitPlan(plan);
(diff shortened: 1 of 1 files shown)


EpsilonPrime (approved):
I argued with myself whether it should be accessible (expecting that folks would need to pull the data out some other way or could decide to enter in a different place).  Glad to have an opinion to help push the design one way or the other!

Related piece of information not useful to this review:  I believe that this visitor class can be fully automatically constructed (useful to keep in line with the proto definition) but it can also be generated for any given proto as well.  I've seen folks asking for generic protobuffer visitors so I know that there is demand.  Now if I can only find the time...

EpsilonPrime:
Apparently I don't have the ability to update the branch and it won't let me approve the PR until the tests pass.  Could you rebase the PR for me please?  Thanks!

Issue #121, opened 2024-09-09
Aggregate relations do not yet support output mapping changes


mortbopet:
Unfortunately, my queries seem to hit the [Aggregate relations do not yet support output mapping changes](https://github.com/substrait-io/substrait-cpp/blob/main/src/substrait/textplan/converter/InitialPlanProtoVisitor.cpp#L634) error. However, for my given usecase, sufficient information is retained/parsed if i just comment out the following lines:
https://github.com/substrait-io/substrait-cpp/blob/03760889013210041038398e595639d04b4f1cc3/src/substrait/textplan/converter/InitialPlanProtoVisitor.cpp#L632-L636

Now that may be bad... but again, works for my usecase. My question is therefore - is there a way to opt-in to this feature, or make the error recoverable, s.t. users (like me) can use upstream substrait-cpp?

EpsilonPrime:
I put that check in there because it's important to handle (it affects the rest of the plan) but I didn't see any cases where folks were using it (in fact, some systems don't even support emits in aggregations yet).  I'll make a stab at adding this missing feature.

mortbopet:
> I put that check in there because it's important to handle (it affects the rest of the plan) but I didn't see any cases where folks were using it (in fact, some systems don't even support emits in aggregations yet). I'll make a stab at adding this missing feature.

Much appreciated! 🙏

EpsilonPrime:
Progress update:  I have binary to text conversion working correctly but there are one or two small bugs in the text to binary parsing that I need to fix.  Shouldn't be more than another day.

EpsilonPrime:
Please let me know if I missed anything.

mortbopet:
So far looks like the fix is working for my use-case. Thank you very much for looking into this!

PR #122, opened 2024-09-12
feat: add support for emit in aggregate relations

EpsilonPrime:
This adds support for emit in aggregations where there no more than one grouping section.

This addresses #121 .


src/substrait/textplan/PlanPrinterVisitor.cpp
```diff
@@ -71,6 +71,19 @@ std::string visitEnumArgument(const std::string& str) {
   return text.str();
 }
 
+bool isAggregate(const SymbolInfo* symbol) {
+  if (symbol->subtype.type() == typeid(::substrait::proto::Rel::RelTypeCase) &&
+      ANY_CAST(::substrait::proto::Rel::RelTypeCase, symbol->subtype) ==
+          ::substrait::proto::Rel::kAggregate) {
+    return true;
+  }
```
bkietz, line 79: For pattern matching like this, I recommend adding a conditional cast macro:

```c++
#define ANY_CAST_IF(ValueType, value) \
  value.type() != typeid(ValueType) \
  ? ::std::nullopt \
  : ::std::make_optional(ANY_CAST(ValueType, value))
```

Then you can write
```suggestion
  if (auto type_case = ANY_CAST_IF(::substrait::proto::Rel::RelTypeCase, symbol->subtype)) {
    return type_case == ::substrait::proto::Rel::kAggregate;
  }
```

If you want to be even fancier, I can provide a `match()` helper:

```c++
bool isAggregate(const SymbolInfo* symbol) {
  return match(symbol->subtype,
    [](::substrait::proto::Rel::RelTypeCase type_case) {
      return type_case == ::substrait::proto::Rel::kAggregate;
    },
    [](RelationType rel_type) {
      return rel_type == RelationType::kAggregate;
    },
    [](...) {
      return false;
    });
}
```

src/substrait/textplan/PlanPrinterVisitor.cpp
```diff
@@ -177,6 +190,65 @@ std::string PlanPrinterVisitor::lookupFieldReference(
...
+      fieldReferencesSize + relationData->generatedFieldReferences.size()) {
+    symbol =
+        relationData
+            ->generatedFieldReferences[fieldReference - fieldReferencesSize];
+  } else {
+    errorListener_->addError(
+        "Encountered field reference out of range: " +
+        std::to_string(fieldReference));
+    return "field#" + std::to_string(fieldReference);
+  }
```
bkietz, line 243: This seems odd to me because I'd expect a clean outer branch on the type of `currentScope` *before* we look at the value of fieldReference. If currentScope *is* an aggregate but fieldReference is out of range, is it really reasonable to use the same code we would have used if currentScope were not an aggregate? I'd expect something like:
```suggestion
  if (isAggregate(currentScope)) {
    if (fieldReference < relationData->generatedFieldReferences.size()) {
      symbol = relationData->generatedFieldReferences[fieldReference];
    }
  } else {
    auto size = relationData->fieldReferences.size();
    if (fieldReference < size) {
      symbol = relationData->fieldReferences[fieldReference];
    } else if (
        fieldReference <
        size + relationData->generatedFieldReferences.size()) {
      symbol =
          relationData
              ->generatedFieldReferences[fieldReference - size];
    }
  }
  if (symbol == nullptr) {
    errorListener_->addError(
        "Encountered field reference out of range: " +
        std::to_string(fieldReference));
    return "field#" + std::to_string(fieldReference);
  }
```

in any case, please clarify the branching here

src/substrait/textplan/converter/InitialPlanProtoVisitor.cpp
```diff
@@ -629,9 +630,24 @@ void InitialPlanProtoVisitor::updateLocalSchema(
...
+      relationData->outputFieldReferences.clear(); // Start over.
+      for (auto item : mapping) {
+        if (item < generatedFieldReferenceSize) {
+          relationData->outputFieldReferences.push_back(
+              relationData->generatedFieldReferences[item]);
+        } else {
+          // TODO -- Add support for grouping fields (needs text syntax).
+          errorListener_->addError(
+              "Asked to emit a field (" + std::to_string(item) +
+              " beyond what the aggregate produced.");
```
bkietz, line 648:
```suggestion
              ") beyond what the aggregate produced.");
```

src/substrait/textplan/parser/SubstraitPlanSubqueryRelationVisitor.cpp
```diff
@@ -374,6 +372,11 @@ comparisonToProto(const std::string& text) {
       Expression_Subquery_SetComparison_ComparisonOp_COMPARISON_OP_UNSPECIFIED;
 }
 
+bool isAggregate(const SymbolInfo* symbol) {
+  return symbol->subtype.type() == typeid(RelationType) &&
+      ANY_CAST(RelationType, symbol->subtype) == RelationType::kAggregate;
```
bkietz, line 377:
```suggestion
  return ANY_CAST_IF(RelationType, symbol->subtype) == RelationType::kAggregate;
```

bkietz (changes requested):

src/substrait/textplan/PlanPrinterVisitor.cpp
EpsilonPrime: ANY_CAST_IF is roughly the amount of tech that I think I need at this point given how often I use ANY_CAST_IF (and I should be fixing things to use zero).  I am much appreciative of the ANY_CAST_IF structure -- definitely is more usable/readable.

src/substrait/textplan/PlanPrinterVisitor.cpp
EpsilonPrime: Agreed, much more understandable with your approach.

src/substrait/textplan/converter/InitialPlanProtoVisitor.cpp
EpsilonPrime: Done.

src/substrait/textplan/parser/SubstraitPlanSubqueryRelationVisitor.cpp
EpsilonPrime: Done.

src/substrait/textplan/converter/InitialPlanProtoVisitor.cpp
```diff
@@ -629,9 +630,24 @@ void InitialPlanProtoVisitor::updateLocalSchema(
   // Revamp the output based on the output mapping if present.
   auto mapping = getOutputMapping(relation);
   if (!mapping.empty()) {
+    // TODO -- Use a more explicit check.
     if (!relationData->outputFieldReferences.empty()) {
```
bkietz, line 634: What sort of more explicit check were you thinking?

```suggestion
    if (!relationData->outputFieldReferences.empty()) {
      assert(relationData->relation.rel_type_case() == ::substrait::proto::Rel::kAggregate));
```

I'm not familiar enough with substrait to know at what level aggregate nodes are the only relations with output field references. Do you mean they are the only ones supported by substrait-cpp? The only relation for which textplan will ever provide a representation? I don't *think* they're the only relations which substrait itself allows to have an output mapping...

I think the comment should explain this in greater depth. If it's the last thing (IE, part of general substrait knowledge) maybe a link to an appropriate section of the substrait doc is in order. 

In any case, if this block of code is specific to `relationData->relation.rel_type_case() == ::substrait::proto::Rel::kAggregate`, it should probably be moved into the switch above

src/substrait/textplan/parser/SubstraitPlanRelationVisitor.cpp
```diff
@@ -316,6 +316,13 @@ bool isRelationEmitDetail(SubstraitPlanParser::Relation_detailContext* ctx) {
       nullptr;
 }
 
+bool isAggregate(const SymbolInfo* symbol) {
```
bkietz, line 319: Why are there two implementations of this function (PlanPrinterVisitor.cpp:74)? The definitions are even different. If the difference is intentional, that's definitely worth a comment or at least renaming the functions to clarify their distinct behaviors. If the difference is not intentional, could this become `SymbolInfo::isAggregate`?

src/substrait/textplan/parser/SubstraitPlanRelationVisitor.cpp
```diff
@@ -2075,10 +2087,19 @@ void SubstraitPlanRelationVisitor::applyOutputMappingToSchema(
   if (common->emit().output_mapping_size() == 0) {
     common->mutable_direct();
   } else {
-    if (!relationData->outputFieldReferences.empty()) {
-      // TODO -- Add support for aggregate relations.
-      errorListener_->addError(
-          token, "Aggregate relations do not yet support emit sections.");
+    if (relationData->relation.has_aggregate()) {
```
bkietz, line 2090: This looks very similar to the code at the end of `InitialPlanProtoVisitor::updateLocalSchema`. Any chance it could be extracted to a helper function?

src/substrait/textplan/parser/SubstraitPlanSubqueryRelationVisitor.cpp
```diff
@@ -374,6 +372,13 @@ comparisonToProto(const std::string& text) {
       Expression_Subquery_SetComparison_ComparisonOp_COMPARISON_OP_UNSPECIFIED;
 }
 
+bool isAggregate(const SymbolInfo* symbol) {
```
bkietz, line 375: ditto

src/substrait/textplan/parser/SubstraitPlanSubqueryRelationVisitor.cpp
```diff
@@ -2234,10 +2244,19 @@ void SubstraitPlanSubqueryRelationVisitor::applyOutputMappingToSchema(
   if (common->emit().output_mapping_size() == 0) {
     common->mutable_direct();
   } else {
-    if (!relationData->outputFieldReferences.empty()) {
-      // TODO -- Add support for aggregate relations.
-      errorListener_->addError(
-          token, "Aggregate relations do not yet support emit sections.");
+    if (relationData->relation.has_aggregate()) {
```
bkietz, line 2247: more ditto

bkietz (changes requested):

src/substrait/textplan/PlanPrinterVisitor.cpp
```diff
@@ -71,6 +71,19 @@ std::string visitEnumArgument(const std::string& str) {
   return text.str();
 }
 
+bool isAggregate(const SymbolInfo* symbol) {
+  if (const auto typeCase =
+          ANY_CAST_IF(::substrait::proto::Rel::RelTypeCase, symbol->subtype)) {
+    if (typeCase == ::substrait::proto::Rel::kAggregate) {
+      return true;
+    }
```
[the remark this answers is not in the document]
EpsilonPrime: Done.

src/substrait/textplan/PlanPrinterVisitor.cpp
```diff
@@ -177,6 +190,66 @@ std::string PlanPrinterVisitor::lookupFieldReference(
...
+        errorListener_->addError(
+            "Internal error: Missing previously encountered parent query symbol.");
+        return "field#" + std::to_string(fieldReference);
+      }
+    }
+  }
+  auto relationData =
+      ANY_CAST(std::shared_ptr<RelationData>, actualScope->blob);
+  const SymbolInfo* symbol{nullptr};
+  if (isAggregate(currentScope)) {
```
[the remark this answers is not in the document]
EpsilonPrime: I do agree that better error messages are useful but I believe the substrait validator is better suited for helping developers figure what is wrong with their plan generation (allowing this code to be simpler).  I've added the type here though.

src/substrait/textplan/parser/SubstraitPlanRelationVisitor.cpp
EpsilonPrime: Long term there shouldn't be this function at all and just a single type in the representation.  Right now the intermediate form can have both types (depending on how it was parsed) which needs to be addressed.

src/substrait/textplan/converter/InitialPlanProtoVisitor.cpp
EpsilonPrime: This was relation into the comment below.  I've replaced the output field references with a check on the incoming proto's type.  I did not move it above as the function is currently separated into a section where the fields are populated with a later emit handling step.  I can see some potential refactoring where those two parts of this rather long function could be split up.

src/substrait/textplan/parser/SubstraitPlanRelationVisitor.cpp
bkietz: In that case, a comment would definitely be advisable just to give future readers a clue that this is a stopgap and that in the future there should only be `RelationType::kAggregate`

bkietz (approved):

richtia (approved):

PR #124, opened 2024-09-23
feat: update to Substrait 0.43, add min. precision timestamp support

ingomueller-net:
Update to Substrait v0.43.0, which is the next version, and add minimal support for precision timestamp support, which is the main change since the previously used v0.42.1. The minimal support is achieved by adding the new type cases to the type switches, where they either produce "not implemented" errors or do the simple functionality of setting nullability of the type, as well as suppressing warnings of using depcreated protobuf accessors in the places where these were not suppressed yet.

### DIFF
--- a/src/substrait/textplan/PlanPrinterVisitor.cpp
+++ b/src/substrait/textplan/PlanPrinterVisitor.cpp
@@ -476,6 +476,8 @@ std::any PlanPrinterVisitor::visitType(const ::substrait::proto::Type& type) {
     case ::substrait::proto::Type::kMap:
     case ::substrait::proto::Type::kUserDefined:
     case ::substrait::proto::Type::kUserDefinedTypeReference:
+    case ::substrait::proto::Type::kPrecisionTimestamp:
+    case ::substrait::proto::Type::kPrecisionTimestampTz:
       errorListener_->addError(
           "Unsupported type requested: " + type.ShortDebugString());
       return std::string("UNSUPPORTED_TYPE");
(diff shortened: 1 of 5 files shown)


ingomueller-net:
I have a couple of subsequent changes that bring this repo up to date with the current version of the main repo and will push them in a bit.

ingomueller-net:
I just realized that the minimal changes are not enough and that the full type implementation is required. I have no intention in doing that. I have started [here](https://github.com/ingomueller-net/substrait-cpp/tree/update-substrait-v0.43.0-fixes-stub), tough, so if someone wants to pick it up, feel free...

ingomueller-net:
> I have a couple of subsequent changes that bring this repo up to date with the current version of the main repo and will push them in a bit.

The remaining changes related to renaming the join types, which is trivial, and introducing another type, which I also have no intention implementing at the moment, so I'll abandon those changes.

EpsilonPrime (approved):
Thanks for the fix!   Adding support for the new join types are on my list.

ingomueller-net:
Thanks, @EpsilonPrime, for approving! Note, though, that there is quite some work left, which I had underestimated when I started. The problem is that the precision timestamps appear in [some function signatures](https://github.com/substrait-io/substrait/blob/main/extensions/functions_datetime.yaml#L86), which are loaded as part of the function tests, which means that these types actually need to be implemented.

ingomueller-net:
One way around that, which may be a good idea anyways, could be to (optionally?) allow loading the function library partially, i.e., to ignore unsupported functions.

PR #125, opened 2024-09-23
fix: use proper string concatenation for error on unsupported types

ingomueller-net:
The previous default template instances of `decodeType` for parametrized types used a "concatenation" on two char pointers, which doesn't work. This PR fixes that by making one of the two an `std::string`. The error did not show up until now because all supported types have specializations of the template.


EpsilonPrime (approved):
Thanks for the fix!

PR #126, opened 2024-10-10
More windows-related fixes

mortbopet:
- use `!` instead of a `not` macro
- Don't build the `planparser` tool on windows; this uses unix commands to parse arguments.
- std::filesystem::path doesn't implicitly convert to std::string on msvc
- A shared library must specify export symbols (or set `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS`) for a .lib file to be generated, which is required for things to link against said library.
- Can't use `*` (wildcard) on windows in a CMake command; use built-in CMake globbing instead.
- substrait-cpp tags an ancient version of protobuf (23.4) which itself uses an ancient version of abseil, that fails to build on MSVC. Bump protobuf to the newest version (28.2).
- various nitty changes related to the protobuf bump

Also requires https://github.com/EpsilonPrime/protobuf-matchers/pull/1/commits/91968ca857e30a07bf1996e195d0f5bf544a2c4e to land.

### DIFF
--- a/export/planloader/CMakeLists.txt
+++ b/export/planloader/CMakeLists.txt
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
+set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
 add_library(planloader SHARED planloader.cpp)
 
 add_dependencies(planloader substrait_io)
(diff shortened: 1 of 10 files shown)


EpsilonPrime (approved):
I'm fine with all these changes.

Of the tools available the plantransformer is probably the only one I truly use regularly (the planparser and planconverter were more useful during development).  Maybe I'll make those optional in the same way tests are.

mortbopet:
@EpsilonPrime Awesome... might need some help with this PR wrt. getting CI to pass. Build/test is passing, but for some reason, `check` is failing.

mbrobbel:
> @EpsilonPrime Awesome... might need some help with this PR wrt. getting CI to pass. Build/test is passing, but for some reason, `check` is failing.

Looks like the runner is getting killed. Maybe it's running out of memory? Can you try to reduce the parallelism here: https://github.com/substrait-io/substrait-cpp/blob/68036ca1209718f597bfe1cd55a7324bdb6f7ae9/scripts/run-clang-tidy.sh#L9?

mortbopet:
@mbrobbel thanks for the tip - reduced to `-j 2`, and things now pass :).

PR #127, opened 2025-02-25
Move ANTLR4 dependency inclusion to top-level CMakeLists.txt

mortbopet:
MSVC has a hard 260 character limit which (unfortunately) isn't immediately fixable (note: this is independent of the "infamous" Windows path length limit).

`substrait-cpp` can become an issue when it itself is being used as a dependency of other projects ([see e.g. this discussion](https://github.com/substrait-io/substrait-mlir-contrib/pull/92#issuecomment-2678199984).

Here, it's the `antlr4` build that's generating the longest paths, e.g.:
> `substrait-cpp/src/substrait/textplan/parser/grammar/antlr4_runtime/src/antlr4_runtime/runtime/Cpp/CMakeFiles/CMakeScratch/TryCompile-s3o9if/CMakeFiles/cmTC_945a8.dir/./`

To give ourselves a bit of breathing room here, `src/substrait/textplan/parser/grammar` seems fairly unnecessary, and is simply a side-effect of `include(ExternalAntlr4Cpp)` being called at that nested location.
This PR moves that inclusion up to the top-level, giving MSVC builds an extra 38 characters to work with 🙂.

Another possible work around could be to rewrite the ANTLR4 dependency management s.t. `FetchContent` is used to manage the build.
This would mean that the dependency would be built in e.g. the root `build/_deps/antlr4` location, irrespective of any git submodule nesting.

* Also allows setting `ANTLR_BUILD_CPP_TESTS` for the ANTLR build (avoids ANTLR pulling in gtest which may conflict with other things when `substrait-cpp` is a dependency of other projects).
* Removes the assumption that antlr libraries are located in the `dist` folder within the ANTLR checked-out source hierarchy (we're doing an in-source build). AFAICT, [this behavior places a hard assumption on `ANTLR_BUILD_CPP_TESTS` being set](https://github.com/antlr/antlr4/blob/dev/runtime/Cpp/runtime/CMakeLists.txt#L159-L177) which is what triggers the behaviour for copying binaries to the `dist` folder.


mortbopet:
cc @EpsilonPrime - looks like the formatting pipelines have false positives (or rather, are checking the entire repo and not just the modified files of this PR).

EpsilonPrime:
I'm okay with removing the formatting checks from .github/workflows/build_test.yml for now.  I suspect the --diff-filter option might solve our issue if we can put the changed files in a separate location but that's outside of the scope of this effort.

EpsilonPrime (approved):

PR #131, opened 2025-02-27
build: suppress warnings in external dependencies

ingomueller-net:
This PR adds compiler flags to the build process of the external dependencies downloaded and built via `FetchContent`, namely GTest and Protobuf. That build process generated more than 2k warnings in my system and, since those are well-tested external libraries, are almost certain all harmless.


ingomueller-net:
The clang-format tests fails even on a file that this PR does not touch. Presumably, #129 fixes that.

EpsilonPrime (approved):

PR #133, opened 2025-02-28
Rename 'build' job to 'build_and_test'

mortbopet:
... because it does both :).


EpsilonPrime (approved):
Thanks for another great fix!

PR #134, opened 2025-02-28
Add windows job to CI build & test

mortbopet:
`find_vs.ps1` is used to enter a visual studio development environment, which in turn, makes `ninja` available.


EpsilonPrime:
I think I updated the require rule to the new name so next time I hope it updates the checks appropriately.

mortbopet:
Currently blocked by https://github.com/antlr/antlr4/pull/4738.

mortbopet:
@EpsilonPrime this PR also bumps the protobuf version to 29.3, required due to [another MSVC issue that removed a definition for `NAN` - which subsequently has been fixed in protobuf at a later tag than what we're currently pointing at](https://github.com/protocolbuffers/protobuf/blob/main/upb/message/internal/message.c#L22-L26).

Let me know if you want that change factored out into a separate PR.

EpsilonPrime:
I'm okay keeping it here if it passes.  The new version doesn't appear to be negatively affecting the non-windows builds.

mortbopet:
@EpsilonPrime Marking this as ready for review. Windows seems to be in a pretty good spot, with _most_ tests passing.

Unfortunately, there's a lingering issue [here](https://github.com/mortbopet/substrait-cpp/blob/main/src/substrait/textplan/parser/SubstraitPlanSubqueryRelationVisitor.cpp) where a `std::any_cast` exception occurs. I've spent a fair bit of time trying to track down the issue, but to no avail. I don't plan on using the textual format any time soon, so for the sake of progress, i've disabled the text-format related tests for now. After merging this PR, I'll create an issue to track it, in case some Windows user stumbles upon this in the future.

EpsilonPrime (approved):
There are definitely some tricks to deal with any_cast errors.

